PR #1, opened 2015-01-20
Initial commit

josenavas:


josenavas:
@ekopylova @amnona The code is ready for review. I'm facing some issues with travis that I'm trying to resolve, but in the meantime you can start reviewing the code.

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+        unaligned.
+    """
+    seqs = [Sequence(label, seq) for label, seq in input_seqs]
+
+    if len(seqs) == 0:
+        raise ValueError("No sequences found!")
+
+    # Check that all the sequence lengths (aligned and unaligned are the same)
+    aligned_lengths = set(s.length for s in seqs)
+    unaligned_lengths = set(s.unaligned_length for s in seqs)
```
wasade: Why should the unaligned sequences be constrained to the same length?

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+    list of Sequence
+
+    Raises
+    ------
+    ValueError
+        If no sequences where found in `input_seqs`
+        If all the sequences do not have the same length either aligned or
+        unaligned.
+    """
+    seqs = [Sequence(label, seq) for label, seq in input_seqs]
```
wasade: This will be prohibitive if run on a HiSeq dataset

wasade: ...though, the order in which git presents the files in the PR may be confusing the situation...

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+    Xi = max frequency of error hamming. The length of this array - 1 limits
+    the hamming distance taken into account, i.e. if the length if `error_dist`
+    is 10, sequences up to 10 - 1 = 9 hamming distance will be taken into
+    account
+    """
+    # Get the sequences
+    seqs = get_sequences(input_seqs)
+
+    # If mean error is not provided, use the same value as read_error
+    mean_error = mean_error if mean_error else read_error
```
wasade: `if mean_error is not None`

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # if error_list not supplied, use the default (22 mock mixture setup)
+    mod_factor = pow((1 - mean_error), seqs[0].unaligned_length)
+    if not error_dist:
```
wasade: `if error_dist is not None`

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+    max_h_dist = len(error_dist)-1
+
+    for seq_i in seqs:
+        # no need to remove neighbors if freq. is <=0
+        if seq_i.frequency <= 0:
+            continue
+
+        # Correct for the fact that many reads are expected to be mutated
+        num_err = error_dist * seq_i.frequency
+        num_err = [val * seq_i.frequency for val in error_dist]
```
wasade: setting `num_err` twice? The one on line 117 is more numpythonic

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # Compare to all other sequences and calculate hamming dist
+        seq_i_len = len(seq_i.sequence.rstrip('-'))
+        for seq_j in seqs:
+            # Ignore current sequence
+            if seq_i == seq_j:
+                continue
+
+            # Calculate the hamming distance
+            h_dist = np.count_nonzero(np.not_equal(seq_i.np_sequence,
```
wasade: if the same length, then this should be equiv to `h_dist = (seq_i.np_sequence != seq_j.np_sequence).sum()`

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+            # Close, so lets calculate exact distance
+            num_substitutions = 0
+            num_indels = 0
+
+            # We stop checking in the shortest sequence after removing trailing
+            # indels. We need to do this in order to avoid double counting
+            # the insertions/deletions
+            l = min(seq_i_len, len(seq_j.sequence.rstrip('-')))
+            for bi, bj in zip(seq_i.np_sequence[:l], seq_i.np_sequence[:l]):
+                if bi != bj:
```
wasade: `bi` and `bj` cannot both be gap, correct?

deblur/deblurring.py
Jorge-C: That also works when sequences don't have the same lenght (i.e., it should be exactly equivalent).

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+
+            # Close, so lets calculate exact distance
+            num_substitutions = 0
+            num_indels = 0
+
+            # We stop checking in the shortest sequence after removing trailing
+            # indels. We need to do this in order to avoid double counting
+            # the insertions/deletions
+            l = min(seq_i_len, len(seq_j.sequence.rstrip('-')))
+            for bi, bj in zip(seq_i.np_sequence[:l], seq_i.np_sequence[:l]):
```
wasade: I believe this forloop can be reduced to the following numpy operations:

```python
14:16:05 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/picrust
> a = array([0,1,1,2,4,4,3,1,2])

14:16:43 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/picrust
> b = array([0,1,4,2,3,4,2,2,2])

14:16:43 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/picrust
> mask = a != b

14:16:53 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/picrust
> muttype = logical_or(where(a[mask] == 4, True, False), where(b[mask] == 4, True, False))

14:17:42 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/picrust
> muttype
Out[26]: array([ True,  True, False, False], dtype=bool)
```

`num_indels` should be the sum of `muttype` and `num_substitutions` should be the length of `muttype` minus its sum.

deblur/sequence.py
```diff
@@ -0,0 +1,49 @@
...
+from collections import defaultdict
+import re
+
+import numpy as np
+
+trans_dict = defaultdict(lambda: 5, A=0, C=1, G=2, T=3)
+trans_dict['-'] = 4
+
+
+class Sequence(object):
```
wasade: Not feasible to use scikit-bio's?

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+    seqs = get_sequences(input_seqs)
+
+    # If mean error is not provided, use the same value as read_error
+    mean_error = mean_error if mean_error else read_error
+
+    # if error_list not supplied, use the default (22 mock mixture setup)
+    mod_factor = pow((1 - mean_error), seqs[0].unaligned_length)
+    if not error_dist:
+        error_dist = np.array(
+            [1.0 / mod_factor, pow(read_error, 1) / mod_factor, 0.01, 0.01,
```
Jorge-C: Is the parenthesis in the right place? Otherwise, the second terms is `read_error**1 == read_error`, so the power is not necessary.

scripts/deblur
```diff
@@ -0,0 +1,57 @@
...
+def deblur_cmds():
+    pass
+
+
+@deblur_cmds.command()
+@click.argument('seqs_fp', required=True,
+                type=click.Path(resolve_path=True, readable=True, exists=True,
+                                file_okay=True))
+@click.option('--read-error', '-e', required=False, type=float, default=0.05,
+              help="Read error rate")
```
ekopylova: The default values for all of these options aren't showing up when calling `deblur deblur_seqs --help`

deblur/sequence.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, label, sequence):
+        self.label = label
+        self.sequence = sequence.upper()
+        self.length = len(self.sequence)
+        self.unaligned_length = self.length - self.sequence.count('-')
+        self.frequency = float(re.search('(?<=size=)\w+', self.label).group(0))
+        self.np_sequence = np.array(
+            [trans_dict[b] for b in self.sequence], dtype=np.int8)
```
wasade: This might be slow... a lookup table is probably faster as it doesn't have the (constant) overhead of computing the hash and it can be done entirely in numpy. Given this is the sequence object, I'm assuming this is a heavy used piece of code. Non-blocking of course, but might be low hanging fruit for a performance boost

deblur/sequence.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __eq__(self, other):
```
wasade: two sequences aren't equal if their IDs differ?

deblur/test/test_deblurring.py
```diff
@@ -0,0 +1,282 @@
+# -----------------------------------------------------------------------------
+# Copyright (c) 2013, The Deblur Development Team.
```
wasade: copyright 2015...

scripts/deblur
```diff
@@ -0,0 +1,57 @@
...
+@click.option('--mean-error', '-m', required=False, type=float, default=None,
+              help="The mean error, used for original sequence estimate. If "
+                   "not passed the same value as --read-error will be used")
+@click.option('--error-dist', '-e', required=False, type=str, default=None,
+              help="A comma separated list of error probabilities for each "
+                   "hamming distance. The length of the list determines the "
+                   "number of hamming distances taken into account.")
+@click.option('--indel-prob', '-i', required=False, type=float, default=0.01,
+              help='Indel probability (same for N indels)')
```
ekopylova: better to explain what is meant by 'indel' .. `help='Insertion/deletion (indel) probability (same for N indels)'`

scripts/deblur
```diff
@@ -0,0 +1,57 @@
...
+    if error_dist:
+        error_dist = list(map(float, error_dist.split(',')))
+
+    with open(seqs_fp, 'U') as f:
+        seqs = deblur(parse_fasta(f), read_error, mean_error, error_dist,
+                      indel_prob, indel_max)
+
+    output_path = "%s.clean" % seqs_fp
+    with open(output_path, 'w') as f:
+        for s in seqs:
```
wasade: It would reduce this a little if `seqs` where an `Alignment` or `SequenceCollection` skbio object

deblur/deblurring.py
```diff
@@ -0,0 +1,167 @@
...
+            if num_indels > indel_max:
+                correction_value = 0
+            elif num_indels > 0:
+                # remove errors due to (PCR?) indels (saw in 22 mock mixture)
+                correction_value = correction_value * indel_prob
+
+            # met all the criteria - so correct the frequency of the neighbor
+            seq_j.frequency -= correction_value
+
+    result = [s for s in seqs if s.frequency > 0]
```
wasade: this loop is effectively done twice, why not have `get_sequences` filter out sequences whose frequency is `<= 0`?

wasade: oh, I see, nm... its frequency is updated within the loop. fun fun

scripts/deblur
```diff
@@ -0,0 +1,57 @@
...
+from deblur.deblurring import deblur
+
+
+@click.group()
+def deblur_cmds():
+    pass
+
+
+@deblur_cmds.command()
+@click.argument('seqs_fp', required=True,
```
ekopylova: add an option for an output directory or filename?

deblur/deblurring.py
Jorge-C: :+1: 

In fact, the `where` calls aren't necessary
```
muttype = logical_or(a[mask] == 4, b[mask] == 4)
```

wasade: ah, damnit, i n00bed it. good catch :)

deblur/sequence.py
ekopylova: a description of this class would be useful ..

deblur/deblurring.py
ekopylova: Should state here that sequences must be in unaligned FASTA format (I assume that's why you're checking their aligned/unaligned lengths on line 46).

scripts/deblur
```diff
@@ -0,0 +1,57 @@
...
+                type=click.Path(resolve_path=True, readable=True, exists=True,
+                                file_okay=True))
+@click.option('--read-error', '-e', required=False, type=float, default=0.05,
+              help="Read error rate")
+@click.option('--mean-error', '-m', required=False, type=float, default=None,
+              help="The mean error, used for original sequence estimate. If "
+                   "not passed the same value as --read-error will be used")
+@click.option('--error-dist', '-e', required=False, type=str, default=None,
```
ekopylova: says here the default is None, whereas the default is actually coded in deblurring.py `# if error_list not supplied, use the default (22 mock mixture setup)`

deblur/deblurring.py
RNAer: This is a restriction the deblurring algorithm imposes. Unfortunately it only works on the sequences of the same length currently.

scripts/deblur
amnona: "-e" is used twice - in --error-dist and in --read-error

scripts/deblur
amnona: Maybe add an option to deblur a single file instead of a directory?

scripts/deblur
```diff
@@ -0,0 +1,57 @@
...
+@click.option('--indel-max', required=False, type=int, default=3,
+              help="Maximal indel number")
+def deblur_seqs(seqs_fp, read_error, mean_error, error_dist, indel_prob,
+                indel_max):
+    """Clean read errors from illumina reads"""
+    # If the user provided an error_dist value, we map it to a list of floats
+    if error_dist:
+        error_dist = list(map(float, error_dist.split(',')))
+
+    with open(seqs_fp, 'U') as f:
```
amnona: Are we assuming all files in the seq_fp directory are sequence files?

wasade: we don't need to care if using skbio's IO registry, which would be more robust here instead of using `parse_fasta` directly

deblur/deblurring.py
josenavas: @wasade The algorithm needs all sequences loaded into memory. Also, at this point of the workflow, the number of sequences as been highly reduced. Amnon has been running some tests and it looks like we should not be concerned about this right now.

@ekopylova done thanks.

deblur/deblurring.py
josenavas: Done

deblur/deblurring.py
josenavas: `if error_dist is None` - fixed!

deblur/deblurring.py
josenavas: Thanks for pointing this out! I was just copying some code over and I forget to check that! Thanks!

deblur/deblurring.py
josenavas: Ups, yeah, I was making more numpythonic code and I forgot to remove the old one hahaha

deblur/deblurring.py
josenavas: You're awesome guys:
```
In [8]: %timeit np.count_nonzero(np.not_equal(a1, a2))
1000000 loops, best of 3: 690 ns per loop

In [9]: %timeit (a1 != a2).sum()
100000 loops, best of 3: 2.62 µs per loop
```

deblur/deblurring.py
josenavas: They can be both gap, but if they both are, the algorithm doesn't care (i.e. it actually looks only for differences)

deblur/deblurring.py
josenavas: Thanks guys. Just a small correction on muttype:

```
muttype = logical_xor(a[mask] == 4, b[mask] == 4)
```

We want the xor because if they're equal we don't need to count them

josenavas: Nevermind, I just re-read the entire code and it actually works with logical_or...

deblur/deblurring.py
amnona: I think 690ns is better than 2.62us ? or am i missing something?

josenavas: Hahahaha awesome call @amnona reading too fast!!

I think I need a lot of coffee today...

deblur/sequence.py
josenavas: I'm unsure on how to do this with a lookup table. Can you provide some guidance on how to do it? The examples that I found do not allow indexing with chars...

deblur/sequence.py
josenavas: @wasade the problem that I have is that I want to precompute some attributes of the Sequence, and right now I don't have an easy way to add that information to the sequence object.

deblur/sequence.py
josenavas: Uhm... I was using it for testing purposes but I agree that given that the object is general it can be misleading.

deblur/test/test_deblurring.py
josenavas: Good catch!

scripts/deblur
josenavas: @amnona this is requiring always a single filepath; i.e. directories are not supported.

scripts/deblur
josenavas: Those defaults are from @amnona 
As I said in the email thread; let's not worry about default values since until we finish the paper and we do a lot of testing we cannot decide which are the best default values.

josenavas: About the values not showing up, that is a click thing, nothing that I can do AFAIK.

scripts/deblur
josenavas: The default for the input is `None`. Note that this default is the actual value that the variable will get, and I need `None` on the function.
@amnona good catch, fixed

scripts/deblur
josenavas: Done

scripts/deblur
josenavas: @amnona note that the command only accepts a single fasta file as input

@wasade the skbio IO returns a BiologicalSequence, I think right now is better to use `parse_fasta`.

deblur/sequence.py
wasade: chars are 8-bit unsigned ints. Light example [here](https://github.com/biocore/scikit-bio/pull/53#discussion-diff-9171472)

deblur/deblurring.py
wasade: Okay, I see, it does get tricky due the pairwise calculation step. What is the expected upper bound for the number of sequences needed to be held in memory?

deblur/sequence.py
josenavas: Thanks @wasade 
I'll create an issue for this. Due to the code sprint I did not have enough time to make all changes here, but the PR needs to go in so I do not block development. Once the code sprint is over, I'll go back here and make all the improvements that you suggested.
Thanks!

deblur/deblurring.py
wasade: Interesting, the operator is more expensive? That is a bit surprising

deblur/sequence.py
wasade: You could either subclass, or just add on the attributes via `setattr`?

Jorge-C: :+1:

scripts/deblur
wasade: What is the issue with `BiologicalSequence`?

PR #17, opened 2015-03-24
Trim seqs

### LINKED ISSUE #9: Implement trim_seqs

josenavas:
Fixes #9

### DIFF
--- a/deblur/workflow.py
+++ b/deblur/workflow.py
@@ -7,9 +7,22 @@
 # -----------------------------------------------------------------------------
 
 
-def trim_seqs(seqs_fp, trim_len):
-    """Step 1: trim FASTA sequences to specified length"""
-    pass
+def trim_seqs(input_seqs, trim_len):
+    """Step 1: trim FASTA sequences to specified length
+
+    Parameters
+    ----------
+    input_seqs : iterable of (str, str)
+        The list of input sequences in (label, sequence) format
+
+    Returns
+    -------
+    Generator of (str, str)
+        The trimmed sequences in (label, sequence) format
+    """
+    for label, seq in input_seqs:
+        if len(seq) >= trim_len:
+            yield label, seq[:trim_len]
 
 
 def dereplicate_seqs(seqs_fp):
--- a/deblur/workflow.py
+++ b/deblur/workflow.py
@@ -14,6 +14,8 @@ def trim_seqs(input_seqs, trim_len):
     ----------
     input_seqs : iterable of (str, str)
         The list of input sequences in (label, sequence) format
+    trim_len : int
+        Sequence trimming length
 
     Returns
     -------
--- a/scripts/deblur
+++ b/scripts/deblur
@@ -74,10 +74,9 @@ def deblur_seqs(seqs_fp, read_error, mean_error, error_dist, indel_prob,
               help="Sequence trim length")
 def trim(seqs_fp, output_fp, trim_length):
     """Step 1: trim FASTA sequences"""
-    with open(seqs_fp, 'U') as in_f:
-        with open(output_fp, 'w') as out_f:
-            for label, seq in trim_seqs(parse_fasta(in_f), trim_length):
-                out_f.write(">%s\n%s\n" % (label, seq))
+    with open(seqs_fp, 'U') as in_f, open(output_fp, 'w') as out_f:
+        for label, seq in trim_seqs(parse_fasta(in_f), trim_length):
+            out_f.write(">%s\n%s\n" % (label, seq))
 
 
 # SEQUENCE DEREPLICATION COMMAND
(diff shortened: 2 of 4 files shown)


scripts/deblur
```diff
@@ -68,18 +68,16 @@ def deblur_seqs(seqs_fp, read_error, mean_error, error_dist, indel_prob,
...
-def trim(seqs_fp, output_fp, trim_len):
+def trim(seqs_fp, output_fp, trim_length):
     """Step 1: trim FASTA sequences"""
-    with open(seqs_fp, 'U') as f:
-        seqs = trim_seqs(parse_fasta(f), trim_len)
-
-    with open(output_fp, 'w') as f:
-        for s in seqs:
-            f.write(s.to_fasta())
+    with open(seqs_fp, 'U') as in_f:
```
wasade: btw, can do the multiple files at once, see [here](http://stackoverflow.com/questions/4617034/how-can-i-open-multiple-files-using-with-open-in-python)

setup.py
```diff
@@ -46,6 +46,7 @@
       extras_require={'test': ["nose >= 0.10.1", "pep8"],
                       'doc': ["Sphinx >= 1.2.2", "sphinx-bootstrap-theme"]},
       install_requires=['click', 'numpy >= 1.7',
-                        'scikit-bio >= 0.2.2, < 0.3.0'],
+                        'scikit-bio >= 0.2.2, < 0.3.0',
+                        'biom-format'],
```
wasade: suggest `biom-format >= 2.1.3, < 2.2.0`

scripts/deblur
josenavas: Nice! I did not know that! Thanks!

setup.py
josenavas: Thanks! Done

scripts/deblur
```diff
@@ -68,18 +68,15 @@ def deblur_seqs(seqs_fp, read_error, mean_error, error_dist, indel_prob,
...
+def trim(seqs_fp, output_fp, trim_length):
     """Step 1: trim FASTA sequences"""
-    with open(seqs_fp, 'U') as f:
-        seqs = trim_seqs(parse_fasta(f), trim_len)
-
-    with open(output_fp, 'w') as f:
-        for s in seqs:
-            f.write(s.to_fasta())
+    with open(seqs_fp, 'U') as in_f, open(output_fp, 'w') as out_f:
+        for label, seq in trim_seqs(parse_fasta(in_f), trim_length):
```
ElDeveloper: Note that `parse_fasta` will be removed in 0.3.0 (I see setup.py delimits the versions, but still), so I would suggest using `skbio.io.read`, perhaps also in conjunction with the `skbio.io.write` function.

Not a blocker, but I thought I would say something :fallen_leaf:

deblur/workflow.py
```diff
@@ -7,9 +7,22 @@
...
-def trim_seqs(seqs_fp, trim_len):
-    """Step 1: trim FASTA sequences to specified length"""
-    pass
+def trim_seqs(input_seqs, trim_len):
+    """Step 1: trim FASTA sequences to specified length
+
+    Parameters
+    ----------
+    input_seqs : iterable of (str, str)
+        The list of input sequences in (label, sequence) format
```
ElDeveloper: The `trim_len` parameter is not described here.

scripts/deblur
josenavas: Agree, currently the skbio.io.read will return a skbio's BiologicalSequence, which will require a bit more though in terms of including it here; so leaving as it is for now until we can explore how the BiologicalSequence can be included here.
However, I think this is raising a concern: we will no longer have a lightweight parser on skbio that simply returns (label, seq) tuples; which some application my need as the object might be an overkill....

deblur/workflow.py
josenavas: Thanks! added

scripts/deblur
ElDeveloper: Got it, seems to be good to leave as is.

If there are performance concerns, then this is something that should
probably be discussed in the skbio mailing list.

wasade: Agree. There may not actually be significant overhead.

scripts/deblur
```diff
@@ -68,18 +68,15 @@ def deblur_seqs(seqs_fp, read_error, mean_error, error_dist, indel_prob,
                 type=click.Path(resolve_path=True, readable=True, exists=True,
                                 file_okay=True))
 @click.argument('output_fp', required=True,
-                type=click.Path(resolve_path=True, readable=True, exists=True,
+                type=click.Path(resolve_path=True, readable=True, exists=False,
                                 file_okay=True))
 @click.option('--trim-length', '-t', required=False, type=int, default=100,
               help="Sequence trim length")
```
ekopylova: add 'show_default=True' to @click.option() so that the default value is displayed during help ..

ekopylova:
@josenavas, the checks are passing, can we merge this?

josenavas:
There are merge conflicts.

ekopylova:
@josenavas could you fix them? They appear to be a quick fix: multiple definitions of the trim_seq function in workflow.py and a library declaration in test_workflow.py

josenavas:
Merge conflicts fixed

PR #31, opened 2016-03-15
Revert "Split input file on samples and serially process each one"

wasade:
Reverts biocore/deblur#30

this merge brings in #29 unexpectedly, which was not ready for merge


josenavas:
I don't think this applies any more - @wasade reopen if that's not the case

PR #32, opened 2016-03-15
Develop

amnona:
Moved the workflow testing data to external fasta files
And created 2 test datasets (s1 is simple, s2 is complicated)


amnona:
ok
it's now passing all the tests :)

deblur/test/data/readme.txt
```diff
@@ -0,0 +1,7 @@
+70_otus.fasta - the 70% representative set from greengenes 13.5
```
ekopylova: not 13.8? Can we put the latest?

deblur/test/data/readme.txt
```diff
@@ -0,0 +1,7 @@
...
+
+seqs_s1.fa - a simulated dataset with 100 reads derived from simset.s1.fasta using art
+seqs_s2.fa - a simulated dataset with 200 reads derived from simset.s2.fasta using grinder, with 10% chimeras, and additionally 35 phix reads
```
ekopylova: is there still the more detailed explanation of how these were created (e.g. actual ART commands and expected sequences that were part of the comment in the code associated to these datasets)

deblur/test/test_workflow.py
```diff
@@ -196,7 +188,7 @@ def test_remove_artifacts_seqs(self):
                         "TCGGCTTTGTAAATCCCTGGGTAAATAGGGT"),
                ("ref6", "TACCCGCAGCTCAAGTGGTGGTCGCTATTATTGAGCCTAAAACGTCCGTAG"
                         "TCGGCTTTGTAAATCCCTGGGTAAATCGGGT")]
-        ref_fp = join(self.working_dir, "ref.fasta")
+        ref_fp = join(self.working_dir, "ref2.fasta")
```
ekopylova: what's the purpose of ref2, ref3, etc. over ref?

josenavas:
I will take a look at this tomorrow (I'm finish finals today)

deblur/test/test_workflow.py
```diff
@@ -576,63 +568,65 @@ def test_build_index_sortmerna(self):
...
-        self.assertEqual(table_obs, table_exp)
+        biom_fp = launch_workflow(seqs_fp, output_fp, read_error, mean_error,
+                                  error_dist, indel_prob, indel_max,
+                                  trim_length, min_size, (ref_fp,),
+                                  ref_db_fp, negate, threads, delim)
+
+        # get the trimmed ground truth sequences
+        with open(origfilename, 'U') as f:
+            orig_seqs = [item[1] for item in parse_fasta(f)]
+        orig_seqs = [item[:trim_length].upper() for item in orig_seqs]
```
ekopylova: why do you convert the strings to upper case in both orig_seqs and outseqs?

deblur/test/test_workflow.py
```diff
@@ -576,63 +568,65 @@ def test_build_index_sortmerna(self):
...
+        table_obs = load_table(biom_fp)
+        outseqs = table_obs.ids(axis='observation')
+        foundall = True
+        for cseq in orig_seqs:
+            if cseq.upper() not in outseqs:
+                foundall = False
+        # test we see all ground truth sequences
+        self.assertTrue(foundall)
+        # and don't see any additional sequences
+        self.assertTrue(len(outseqs) == len(orig_seqs))
```
ekopylova: could you just use ```self.assertListEquals(outseqs, orig_seqs)``` (should replace both ```self.assertTrue(foundall)``` and ```self.assertTrue(len(outseqs) == len(orig_seqs))```)

ekopylova:
@amnona a few comments, otherwise looks good to me

deblur/test/test_workflow.py
josenavas: You can use `self.assertItemsEqual(outseqs, orig_seqs)` and no need to worry about the sort.

deblur/test/data/readme.txt
amnona: fixed to gg 13.8 (same data since it's just different taxonomy info)

On Thu, Mar 17, 2016 at 4:41 AM, Evguenia Kopylova <notifications@github.com

deblur/test/data/readme.txt
amnona: added code to generate the simulated datasets to the readme

On Thu, Mar 17, 2016 at 4:43 AM, Evguenia Kopylova <notifications@github.com

deblur/test/test_workflow.py
amnona: just to enable debugging (so we know which file is from which stage...)

On Thu, Mar 17, 2016 at 4:45 AM, Evguenia Kopylova <notifications@github.com

deblur/test/test_workflow.py
amnona: since in general simulated reads can have lowercase, and we don't require
anything about case from the deblurring, so to be on the safe side...

On Thu, Mar 17, 2016 at 4:49 AM, Evguenia Kopylova <notifications@github.com

deblur/deblurring.py
```diff
@@ -52,7 +52,7 @@ def get_sequences(input_seqs):
     return seqs
 
 
-def deblur(input_seqs, read_error=0.05, mean_error=None, error_dist=None,
+def deblur(input_seqs, read_error=0.05, mean_error=0.005, error_dist=None,
```
josenavas: @amnona can you explain the rationale behind this change? I though that if the user doesn't provide a `mean_error` we were gonna use the same value as `read_error`. Did that change based on recent benchmarks?

deblur/deblurring.py
```diff
@@ -148,7 +149,7 @@ def deblur(input_seqs, read_error=0.05, mean_error=None, error_dist=None,
             mask = (sub_seq_i != sub_seq_j)
             muttype = np.logical_or(sub_seq_i[mask] == 4, sub_seq_j[mask] == 4)
             num_indels = muttype.sum()
-            num_substitutions = len(muttype) - num_indels
+            num_substitutions = h_dist - num_indels
```
josenavas: I don't think this change is correct - and if it is, then the name of the variables is misleading. Currently this code is doing the hamming distance minus the num of indels. Isn't that conceptually different that counting the number of substitutions as it was done before?

deblur/deblurring.py
```diff
@@ -161,6 +162,6 @@ def deblur(input_seqs, read_error=0.05, mean_error=None, error_dist=None,
             # met all the criteria - so correct the frequency of the neighbor
             seq_j.frequency -= correction_value
 
-    result = [s for s in seqs if s.frequency > 0]
+    result = [s for s in seqs if round(s.frequency) > 0]
```
josenavas: Why the rounding? Benchmarks supporting this?

deblur/test/test_workflow.py
josenavas: Does this mean that those files where never removed? Note that the tests should clean up the environment after running the tests - so ref_fp should be added to the `files_to_remove` list.

deblur/test/test_workflow.py
```diff
@@ -576,63 +568,60 @@ def test_build_index_sortmerna(self):
         files_to_remove_act = [basename(f) for f in files_to_remove]
         self.assertListEqual(files_to_remove_exp, files_to_remove_act)
 
-    def test_launch_workflow(self):
-        """Test launching complete workflow using simulated sequences 1.
+    def run_workflow_try(self, simfilename, origfilename, ref_fp, ref_db_fp):
```
josenavas: Please, change back the name of the function. If it doesn't start with `test_` is not run by nosetests.

josenavas: oh nvm - this is an aux function used in other tests. Can you add the `_` at the beginning to show that is a private function - will prevent confusions as I had.

deblur/test/test_workflow.py
```diff
@@ -576,63 +568,60 @@ def test_build_index_sortmerna(self):
...
+        """Test launching the complete workflow using simulated sequences
+        and compare to original ground truth.
+
+        input:
```
josenavas: This is not following the numpy standards.

Change `input:` to:
```
Parameters
---------------
```

and `string` to `str`

josenavas:
Done with the review. I'm having the concern here that the tests added here are too big to have been run "by hand", so I'm wondering if we may still be missing some bugs - as happened before.

deblur/deblurring.py
amnona: the read error is the upper bound on the error rate, whereas the mean error is the mean. They should not be the same. It was a mistake to make them identical (when i run deblurring i don't use the default values, so didn't notice it until now...)

deblur/deblurring.py
amnona: We need to throw away sequences that are close to 0 (since at the end the read data is discrete). so we get read of sequences that after deblurring have <0.5 read left. Note that otherwise we would get 0 in the seq.to_fasta

RNAer:
:+1:  looks good. will merge it.

PR #34, opened 2016-06-09
new version

amnona:
Fixed bugs (such as failing if a fasta file was empty)
Added logging
Replaced the merge bioms with a per sample fasta file merging into a single biom table (much faster)
Fixed the frequencies following deblur to be correct (and not taken from original file)
Added option to deblur a directory of fasta files (so don't need to redo split again)


deblur/workflow.py
```diff
@@ -61,17 +62,25 @@ def dereplicate_seqs(seqs_fp,
...
+
+    log_name = "%s.log" % output_fp
 
-    vsearch_dereplicate_exact_seqs(
-        fasta_filepath=seqs_fp,
-        output_filepath=output_fp,
-        output_uc=uc_output,
-        minuniquesize=min_size,
-        log_name=log_name)
+    params = ['vsearch', '--derep_fulllength', seqs_fp, '--output', output_fp, '--sizeout']
```
josenavas: Instead of using extend, can you create the entire list at once?

deblur/workflow.py
```diff
@@ -61,17 +62,25 @@ def dereplicate_seqs(seqs_fp,
...
+    params.extend(['--fasta_width', '0', '--threads', '1', '--minuniquesize', str(min_size)])
+    params.extend(['--quiet'])
+    if use_log:
+        params.extend(['--log', log_name])
+    res = subprocess.call(params)
```
josenavas: Instead of using subprocess call, can you use this function from qiita? https://github.com/biocore/qiita/blob/master/qiita_db/processing_job.py#L25

You can just copy/paste that function and use it here directly (it is BSD licensed so no need for extra permission as long as we put a note saying that we copied form Qiita). The advantages of it is that it returns the std_err and std_out, which you can include in the logger.debug/logger.error calls which will help for debugging.

deblur/workflow.py
```diff
@@ -82,27 +91,31 @@ def build_index_sortmerna(ref_fp, working_dir):
...
-                output_dir=working_dir)
-        all_db.append(sortmerna_db)
-        all_files_to_remove.extend(files_to_remove)
-    return tuple(all_db), all_files_to_remove
+        fasta_dir, fasta_filename = split(db)
+        index_basename = splitext(fasta_filename)[0]
+        db_output = join(working_dir, index_basename)
+        logger.debug('processing file %s into location %s' % (db, db_output))
+        params = ['indexdb_rna', '--ref', '%s,%s' % (db, db_output), '--tmpdir', working_dir]
+        res = subprocess.call(params)
```
josenavas: Same comment as above (applies to all - change all `subprocess.call` to `system_call`.

deblur/workflow.py
```diff
@@ -82,27 +91,31 @@ def build_index_sortmerna(ref_fp, working_dir):
...
+        if not res == 0:
+            logger.error('Problem running indexdb_rna on file %s to dir %s. database not indexed' % (db, db_output))
+            continue
```
josenavas: I don't think a continue is correct in here. If there is a problem generating the database, the execution should stop, as the subsequent calls to sortmerna will fail

deblur/workflow.py
```diff
@@ -132,33 +145,41 @@ def remove_artifacts_seqs(seqs_fp,
...
+        params.extend(['--blast', '3', '--best', '1', '--print_all_reads'])
+        params.extend(['-v'])
+
+        with open(blast_output + '.log', "w") as f:
+            res = subprocess.call(params, stdout=f)
+        if not res == 0:
+            logger.critical('sortmerna error on file %s' % seqs_fp)
+            return output_fp
+
+        bfl = open(blast_output+'.blast', 'r')
```
josenavas: use `with open('%s.blast' % bast_output. 'r') as bfl:`
and indent the code below - you can then remove the `bfl.close()` call on line 182

deblur/workflow.py
```diff
@@ -189,13 +219,31 @@ def multiple_sequence_alignment(seqs_fp, threads=1):
     Returns
     -------
     Alignment object
-        The aligned sequences.
+        The aligned sequences. False if file does not exist
```
josenavas: This is not good practice - it should return None

deblur/workflow.py
```diff
@@ -189,13 +219,31 @@ def multiple_sequence_alignment(seqs_fp, threads=1):
...
     """
-    return align_unaligned_seqs(seqs_fp=seqs_fp, params={'--thread': threads})
+    logger = logging.getLogger(__name__)
+    logger.debug('multiple_sequence_alignment seqs file %s' % seqs_fp)
+
+    if stat(seqs_fp).st_size == 0:
+        logger.info('msa failed. file %s has no reads' % seqs_fp)
+        return False
+    try:
+        #        aligned_seqs = align_unaligned_seqs(seqs_fp=seqs_fp,
```
josenavas: Remove commented code

deblur/workflow.py
```diff
@@ -207,62 +255,32 @@ def remove_chimeras_denovo_from_seqs(seqs_fp, working_dir):
...
+                 'to working dir %s' % (seqs_fp, working_dir))
+
+    output_fp = join(
+        working_dir, "%s.no_chimeras" % basename(seqs_fp))
+
+    # we use the parameters dn=0.000001, xn=1000, minh=10000000
+    # so 1 mismatch in the A/B region will cancel it being labeled as chimera
+    # and ~3 unique reads in each region will make it a chimera if
+    # no mismatches
+    params = ['vsearch', '--uchime_denovo', seqs_fp]
```
josenavas: Same comment as above - create the entire list at once rather than use extend.

deblur/test/test_workflow.py
```diff
@@ -247,17 +246,13 @@ def test_remove_artifacts_seqs_index_prebuilt(self):
...
         self.files_to_remove.append(ref_fp)
         # build index
-        sortmerna_db, files_to_remove = \
-            build_database_sortmerna(
-                fasta_path=ref_fp,
-                max_pos=10000,
-                output_dir=self.working_dir)
-        self.files_to_remove.extend(files_to_remove)
+        sortmerna_db = build_index_sortmerna([ref_fp], self.working_dir)
+#        self.files_to_remove.extend(files_to_remove)
```
josenavas: Remove commented code

deblur/workflow.py
```diff
@@ -61,17 +62,25 @@ def dereplicate_seqs(seqs_fp,
...
+    log_name = "%s.log" % output_fp
 
-    vsearch_dereplicate_exact_seqs(
-        fasta_filepath=seqs_fp,
-        output_filepath=output_fp,
-        output_uc=uc_output,
-        minuniquesize=min_size,
-        log_name=log_name)
+    params = ['vsearch', '--derep_fulllength', seqs_fp, '--output', output_fp, '--sizeout']
+    params.extend(['--fasta_width', '0', '--threads', '1', '--minuniquesize', str(min_size)])
```
josenavas: Why is this hardcoded to 1 thread?

deblur/workflow.py
```diff
@@ -207,62 +255,32 @@ def remove_chimeras_denovo_from_seqs(seqs_fp, working_dir):
...
-            clusters[seq2].extend(derep_clusters[cluster_id])
-    return clusters
+    logger = logging.getLogger(__name__)
+    logger.debug('remove_chimeras_denovo_from_seqs seqs file %s'
+                 'to working dir %s' % (seqs_fp, working_dir))
+
+    output_fp = join(
+        working_dir, "%s.no_chimeras" % basename(seqs_fp))
+
+    # we use the parameters dn=0.000001, xn=1000, minh=10000000
```
josenavas: Is there any case in which the user will want to change these parameters?

deblur/workflow.py
```diff
@@ -207,62 +255,32 @@ def remove_chimeras_denovo_from_seqs(seqs_fp, working_dir):
...
+    # so 1 mismatch in the A/B region will cancel it being labeled as chimera
+    # and ~3 unique reads in each region will make it a chimera if
+    # no mismatches
+    params = ['vsearch', '--uchime_denovo', seqs_fp]
+    params.extend(['--nonchimeras', output_fp])
+    params.extend(['-dn', '0.000001', '-xn', '1000'])
+    params.extend(['-minh', '10000000', '--mindiffs', '5'])
+    params.extend(['--fasta_width', '0'])
+    params.extend(['--threads', '1'])
```
josenavas: This is hardcoded to 1 - why?

deblur/workflow.py
```diff
@@ -401,12 +396,101 @@ def merge_otu_tables(output_fp, all_tables):
...
+    """
+    logger = logging.getLogger(__name__)
+    logger.debug('create_otu_table for %d samples, '
+                 'into output table %s' % (len(deblurred_list), output_fp))
+
+    seqdict = {}
+    seqlist = []
+    sampdict = {}
+    samplist = []
+    obs = scipy.sparse.dok_matrix((1E9, 1E6), dtype=np.int)
```
josenavas: Why is this harcoded to `1E9` and `1E6`?

deblur/workflow.py
```diff
@@ -401,12 +396,101 @@ def merge_otu_tables(output_fp, all_tables):
...
+    for (cfilename, csampleid) in deblurred_list:
+        if csampleid in sampdict:
```
josenavas: If this is an error, why the execution continues rather than actually stopping and raising the error?

deblur/workflow.py
```diff
@@ -401,12 +396,101 @@ def merge_otu_tables(output_fp, all_tables):
...
+            logger.error('sample %s already in table!' % csampleid)
+            continue
+        sampdict[csampleid] = len(sampdict)-1
+        samplist.append(csampleid)
+        csampidx = len(sampdict)-1
+        for chead, cseq in parse_fasta(open(cfilename, 'U')):
```
josenavas: use `with open(cfilename, 'U') as f`...

deblur/workflow.py
```diff
@@ -401,12 +396,101 @@ def merge_otu_tables(output_fp, all_tables):
...
+        if len(cfile) < len(file_end):
+            continue
+        if cfile[-len(file_end):] == file_end:
+            sampleid = cfile[:-len(file_end)]
+            names.append((cname, sampleid))
+    logger.debug('found %d files' % len(names))
+    return names
+
+
+def create_otu_table(output_fp, deblurred_list):
```
josenavas: Can you add comments on this function as well as blank lines when conceptual blocks are done? It is hard to follow what is going on in this function.

deblur/workflow.py
```diff
@@ -460,22 +547,33 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
...
-                     uc_output=True)
+                     min_size=min_size)
     # Step 3: Remove artifacts
     output_artif_fp = remove_artifacts_seqs(seqs_fp=output_derep_fp,
                                             ref_fp=ref_fp,
                                             working_dir=working_dir,
                                             ref_db_fp=ref_db_fp,
                                             negate=negate,
                                             threads=threads)
+    if not output_artif_fp:
```
josenavas: Raise an error

deblur/workflow.py
```diff
@@ -460,22 +547,33 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
...
+        return
     # Step 4: Multiple sequence alignment
     output_msa_fp = join(working_dir,
                          "%s.msa" % basename(output_artif_fp))
-    with open(output_msa_fp, 'w') as f:
-        alignment = multiple_sequence_alignment(seqs_fp=output_artif_fp,
-                                                threads=threads)
-        f.write(alignment.to_fasta())
+    # with open(output_msa_fp, 'w') as f:
+    #     alignment = multiple_sequence_alignment(seqs_fp=output_artif_fp,
```
josenavas: Remove commented code.

deblur/workflow.py
```diff
@@ -460,22 +547,33 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
...
+    #                                             threads=threads)
+    #     if not alignment:
+    #         logger.debug('msa failed. aborting')
+    #         return False
+    #     f.write(alignment.to_fasta())
+    alignment = multiple_sequence_alignment(seqs_fp=output_artif_fp,
+                                            threads=threads)
+    if not alignment:
```
josenavas: Same as before - raise an error after writing in the log.

josenavas:
Done with review

deblur/workflow.py
amnona: This is just to have a large enough sparse matrix (enough samples/otus) so the indices won't be out of bounds. Since the matrix is DOK sparse, these numbers don't affect the memory/speed.
we just assume no more than 1E9 different bacteria and 1E6 samples (can make it larger...but don't think we'll ever reach it...). Anyway this is not a parameter for the user.

deblur/workflow.py
amnona: Some samples can fail the MSA (e.g. when they have 1 read). In this case, we want to just fail on this sample and not stop the pipeline for the rest of the samples (just not include this sample in the final deblurred biom table). I think raising an error will stop the whole deblur script?

deblur/workflow.py
josenavas: I see. However, in the case in which this limit is reached, deblur will fail and it will be a hard limitation of the software. IMO, this is not acceptable. Also, you know the maximum number of samples that you are gonna add, its `len(deblurred_list)`, correct? You can then initialize the matrix as:
```python
obs = scipy.sparse.dok_matrix((1E9, len(deblurred_list), dtype=np.int)
```

Then, for overcoming the limitation on the number of observations, you can change line 480 to extend the matrix once the limit is reached:

```python
try:
    obs[cseqid, csampidx] = cfreq
except IndexError:
    shape = obs.shape
    obs.resize((shape[0]*2,  shape[1]))
    obs[cseqid, csampidx] = cfreq
```
Note that I'm not changing the `shape[1]` value because we have already initialize it to the correct value. The doubling in size is standard CS approach to solve this issue.

This code will not fail even if we have more than 1E9 observations that, hey who knows? we may reach at some point, specially if working with really diverse environments.

deblur/workflow.py
josenavas: The return `False` that you have here will stop the execution of this entire `launch_workflow` instance. Is `launch_workflow` supposed to be executed in a per sample basis? If so it is not clear in the code and probably hard to see with the rest of the changes...

josenavas: In any case, if you don't want to stop the execution of the entire script, you can raise a warning:
```python
from warnings import warn
warn("MSA failed for file: %s" % seqs_fp, UserWarning)
```
Users only look at the log files if the execution failed, so putting something directly on the screen makes them aware that for a given file, no deblur output is generated for whatever reason (I'm assuming that there will be something in the log file that will help the user to know where and why it fail).

deblur/workflow.py
```diff
@@ -401,12 +396,101 @@ def merge_otu_tables(output_fp, all_tables):
...
+                 'file-ending %s' % (input_dir, file_end))
+
+    names = []
+    for cfile in listdir(input_dir):
+        cname = join(input_dir, cfile)
+        if not isfile(cname):
+            continue
+        if len(cfile) < len(file_end):
+            continue
+        if cfile[-len(file_end):] == file_end:
```
RNAer: you can use `cfile.startswith(file_end)` and remove the previous `if` at line 438

deblur/workflow.py
```diff
@@ -401,12 +396,101 @@ def merge_otu_tables(output_fp, all_tables):
...
+            continue
+        sampdict[csampleid] = len(sampdict)-1
+        samplist.append(csampleid)
+        csampidx = len(sampdict)-1
+        for chead, cseq in parse_fasta(open(cfilename, 'U')):
+            if cseq not in seqdict:
+                seqdict[cseq] = len(seqlist)
+                seqlist.append(cseq)
+            cseqidx = seqdict[cseq]
+            cfreq = float(re.search('(?<=size=)\w+', chead).group(0))
```
RNAer: this can be a simple split.

deblur/workflow.py
```diff
@@ -401,12 +396,101 @@ def merge_otu_tables(output_fp, all_tables):
...
+    sampdict = {}
+    samplist = []
+    obs = scipy.sparse.dok_matrix((1E9, 1E6), dtype=np.int)
+    for (cfilename, csampleid) in deblurred_list:
+        if csampleid in sampdict:
+            logger.error('sample %s already in table!' % csampleid)
+            continue
+        sampdict[csampleid] = len(sampdict)-1
+        samplist.append(csampleid)
+        csampidx = len(sampdict)-1
```
RNAer: better to replace `len(sampdict)-1` with `sampdict[csampleid]`?

deblur/workflow.py
```diff
@@ -132,33 +145,41 @@ def remove_artifacts_seqs(seqs_fp,
...
-                with open(stderr_fp, 'U') as stderr_f:
-                    for line in stderr_f:
-                        print(line)
-            raise ValueError("Could not run SortMeRNA.")
-
-        for line in app_result['BlastAlignments']:
+        params = ['sortmerna', '--reads', seqs_fp, '--ref', '%s,%s' % (db, ref_db_fp[i])]
+        params.extend(['--aligned', blast_output])
+        params.extend(['--blast', '3', '--best', '1', '--print_all_reads'])
+        params.extend(['-v'])
```
RNAer: create the entire list.

deblur/workflow.py
wasade: +1 if easy to have rational bounds go avoid hard coding

Btw, didn't realize you were using scipy dok. Awesome! Thought the earlier
benching was with Python dict.

deblur/workflow.py
amnona: added warning to the user, but want to be as tolerant as possible here - just ignore the sample replicate...

deblur/workflow.py
amnona: we don't want to abort the whole pipeline if one sample fails (can happen if sample had 1 read, etc.).
Added a warning to the user instead

deblur/workflow.py
amnona: launch_workflow is executed on a per-sample basis.
clarified in the function definition

deblur/workflow.py
amnona: i think this is a bit more on the safe side (i.e. if the sequence header has "size" in it also in the sample name?
but open to suggestions :)
anyway, this regexp should not slow the program a lot...

deblur/workflow.py
amnona: changed the dict to contain "True". We only use the dict to keep track of which samples we've already processed.

josenavas: In that case, why don't you use a `set`?

deblur/workflow.py
josenavas: I think using the regex is ok - however, it may be useful to first compile the regex outside the for loop (see re.compile) and then just use match/search in here.

amnona:
Fixed the PR comments and added better logging (with user defined log level and filename parameters)

amnona:
also added different thresholds for positive (keep all seqs close enough to database) and negative (remove all seqs very similar to database) artifact removal

deblur/workflow.py
```diff
@@ -61,17 +62,29 @@ def dereplicate_seqs(seqs_fp,
     min_size : integer, optional
         discard sequences with an abundance value smaller
         than integer
-    uc_output: boolean, optional
-        output the dereplication map in .uc format
+    use_log: boolean, optional
+        save the vsearch logfile as well (to output_fp.log)
+        default=False
+    threads : int
```
josenavas:
```
threads : int, optional
    number of threads to use (0 for all available). Default: 1.
```

deblur/workflow.py
```diff
@@ -61,17 +62,29 @@ def dereplicate_seqs(seqs_fp,
...
+    log_name = "%s.log" % output_fp
 
-    vsearch_dereplicate_exact_seqs(
-        fasta_filepath=seqs_fp,
-        output_filepath=output_fp,
-        output_uc=uc_output,
-        minuniquesize=min_size,
-        log_name=log_name)
+    params = ['vsearch', '--derep_fulllength', seqs_fp, '--output', output_fp, '--sizeout',
+              '--fasta_width', '0', '--threads', '1', '--minuniquesize', str(min_size),
```
josenavas: threads is repeated here (once hardcoding to 1 and another one parametrized)

deblur/workflow.py
```diff
@@ -131,34 +152,59 @@ def remove_artifacts_seqs(seqs_fp,
         number of threads to use for SortMeRNA
     verbose: boolean, optional
         If true, output SortMeRNA errors
+    sim_thresh: float, optional
```
josenavas: Missing the description of `coverage_thresh`.

Also, in this case, instead of giving the default as 0, is better to have `None` so the description is something like:
`If not provided, 0.65 is used if negate=False and 0.95 for negate=True`.

josenavas: It will probably also be good to mention that the value should be between 0 and 1.

deblur/workflow.py
josenavas: @amnona this comment has not been addressed (it applies to the function `create_otu_table`

amnona: i think it didn't show because this file failed the flake8... fixed and
re-pushed...

deblur/workflow.py
```diff
@@ -382,40 +347,122 @@ def write_biom_table(table, biom_fp):
...
+                    seqlist.append(cseq)
+                cseqidx = seqdict[cseq]
+                cfreq = float(sizeregexp.search(chead).group(0))
+                try:
+                    obs[cseqidx, csampidx] = cfreq
+                except IndexError:
+                    shape = obs.shape
+                    obs.resize((shape[0]*2,  shape[1]))
+                    obs[cseqidx, csampidx] = cfreq
+    logger.debug('loaded %d samples, %d unique sequences'
```
josenavas: It still doesn't seem to appear. An example of what I'm requesting is, for example, a blank line above this line, so it actually visually breaks the different parts of the code.

deblur/workflow.py
```diff
@@ -438,16 +485,22 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
         filepath(s) to SortMeRNA indexed database for artifact removal
     negate: boolean
         discard all sequences aligning to the ref_fp database
-    threads: integer, optional
-        number of threads to use for SortMeRNA
+    threads_per_sample: integer, optional
```
josenavas: I don't think you need to change the parameter name here - if launch workflow is already running in a per sample basis, is already understood that this 'threads' parameter only applies in a per sample basis.

deblur/workflow.py
```diff
@@ -438,16 +485,22 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
...
+        number of threads to use for SortMeRNA/mafft/vsearch (0 for max available)
     delim: string, optional
         delimiter in FASTA labels to separate sample ID from sequence ID
+    sim_thresh: float, optional
```
josenavas: Same comments as before regarding using None instead of 0.  Also, shouldn't the `coverage_thresh` also be added here?

deblur/workflow.py
```diff
@@ -488,16 +548,69 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
...
+    return output_no_chimeras_fp
+
+
+def start_log(level=logging.DEBUG, filename=''):
+    """
+    start the logger for the run
+
+    Parameters
+    ----------
+    levele : logging.DEBUG, logging.INFO etc. for the log level.
```
josenavas: `levele` -> `level`

deblur/workflow.py
```diff
@@ -488,16 +548,69 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
...
-
-    return biom_fp
+        output_deblur_fp, working_dir, threads=threads_per_sample)
+    logger.info('finished processing file')
+    return output_no_chimeras_fp
+
+
+def start_log(level=logging.DEBUG, filename=''):
+    """
+    start the logger for the run
```
josenavas: Move this line to follow the triple quotes in the line above

deblur/workflow.py
```diff
@@ -488,16 +548,69 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
...
+
+    Parameters
+    ----------
+    levele : logging.DEBUG, logging.INFO etc. for the log level.
+    filename : str
```
josenavas: Use `None`

deblur/workflow.py
josenavas: Also, it is missing the `optional` keywrod and the default in the description. Also the description should be in the next line

deblur/workflow.py
josenavas: Sorry, use `None` as default, rather than the empty string. Also missing the optional keyword

scripts/deblur
```diff
@@ -12,37 +12,42 @@ import click
...
                              split_sequence_file_on_sample_ids_to_files,
-                             merge_otu_tables,
-                             build_index_sortmerna)
+                             build_index_sortmerna,
+                             get_files_for_table,
+                             create_otu_table,
+                             start_log)
+
+
+__version__ = "0.9.2.1"
```
josenavas: Why has this been added?

scripts/deblur
```diff
@@ -52,9 +57,19 @@ def deblur_cmds():
                    '(same for N indels)')
 @click.option('--indel-max', required=False, type=int, default=3,
               help="Maximal indel number")
+@click.option('--log-level', required=False, type=int, default=2,
```
josenavas: Use the range type: http://click.pocoo.org/5/options/#range-options

josenavas: It it applies to other parameters (e.g. mean error and read error) please change.

scripts/deblur
```diff
@@ -79,8 +94,17 @@ def deblur_seqs(seqs_fp, read_error, mean_error, error_dist, indel_prob,
                                 file_okay=True))
 @click.option('--trim-length', '-t', required=False, type=int, default=100,
               help="Sequence trim length")
-def trim(seqs_fp, output_fp, trim_length):
+@click.option('--log-level', required=False, type=int, default=2,
```
josenavas: Same as before (I'm gonna stop commenting this, but please fix in all)

scripts/deblur
```diff
@@ -134,34 +161,85 @@ def dereplicate(seqs_fp, output_fp, min_size, uc_output):
...
               show_default=True, type=bool,
               help="Discard all sequences aligning to the database "
                    "passed via --ref-fp option")
-@click.option('--threads', '-t', required=False, type=int,
+@click.option('--threads-per-sample', '-t', required=False, type=int,
               default=1, show_default=True,
-              help="Number of threads to use for SortMeRNA")
-def remove_artifacts(seqs_fp, output_fp, ref_fp, ref_db_fp,
-                     negate, threads):
+              help="Number of threads to use per sample")
```
josenavas: Add: `Pass 0 to use all available cores`

scripts/deblur
```diff
@@ -188,55 +276,65 @@ def multiple_seq_alignment(seqs_fp, output_fp):
...
 
 # LAUNCH FULL DEBLUR PIPELINE COMMAND
 @deblur_cmds.command()
 @click.option('--seqs-fp', required=True,
               type=click.Path(resolve_path=True, readable=True, exists=True,
-                              file_okay=True),
-              help="Demultiplexed FASTA file including all samples")
+                              file_okay=True, dir_okay=True),
+              help="Demultiplexed FASTA file including all samples or a directory of .fasta files (requires --no-split)")
+@click.option('--no-split', required=False, type=bool, default=False, is_flag=True,
```
josenavas: Do you really need this flag? What I would is check the type of the input using `os.path.isdir`. If it is a directory, then you assume that it is a directory of fasta files, if not, it is a fasta file that you need to split.

scripts/deblur
```diff
@@ -304,26 +418,38 @@ def workflow(seqs_fp, output_dir, ref_fp, ref_db_fp, overwrite,
...
     makedirs(working_dir)
+
     # Split demultiplexed FASTA on sample IDs
-    out_dir_split = join(output_dir, "split")
-    mkdir(out_dir_split)
-    with open(seqs_fp, 'U') as seqs_f:
-        split_sequence_file_on_sample_ids_to_files(
-            seqs_f,
-            out_dir_split)
+    if not no_split:
```
josenavas:
```python
from os.path import isdir  #should go with all the other imports

if isdir:
    ...
```

josenavas:
Done with another review.

deblur/workflow.py
amnona: since the parallelization will be using threads, i prefer to keep this name to prevent confusion between threads for parallel sample processing and threads per sample.

scripts/deblur
amnona: we enable the users to run only these parts of the pipeline as well (i.e. just to join fasta files to a single biom table)

scripts/deblur
amnona: i think there is only int range in click, and errors are floats...

amnona:
fixed more PR issues.

deblur/workflow.py
```diff
@@ -82,27 +95,33 @@ def build_index_sortmerna(ref_fp, working_dir):
...
+        db_output = join(working_dir, index_basename)
+        logger.debug('processing file %s into location %s' % (db, db_output))
+        params = ['indexdb_rna', '--ref', '%s,%s' % (db, db_output), '--tmpdir', working_dir]
+        sout, serr, res = _system_call(params)
+        if not res == 0:
+            logger.error('Problem running indexdb_rna on file %s to dir %s. database not indexed' % (db, db_output))
+            logger.debug('stdout: %s' % sout)
+            logger.debug('stderr: %s' % serr)
+            logger.critical('execution halted')
+            raise Exception('Cannot index database file %s' % db)
```
wasade: Could this be a `RuntimeError` please?

deblur/workflow.py
```diff
@@ -167,12 +217,21 @@ def op(x): return x in aligned_seq_ids
 
     # if negate = False, only output sequences
     # matching to at least one of the databases
+    totalseqs = 0
+    okseqs = 0
+    badseqs = 0
     with open(seqs_fp, 'U') as seqs_f:
```
wasade: these nested contexts can be collapsed to:

```python
with open(seqs_fp, 'U') as seqs_f, open(output_fp, 'w') as out_f:
    ...
```

deblur/workflow.py
```diff
@@ -167,12 +217,21 @@ def op(x): return x in aligned_seq_ids
...
         with open(output_fp, 'w') as out_f:
             for label, seq in parse_fasta(seqs_f):
+                totalseqs += 1
                 label = label.split()[0]
                 if op(label):
                         out_f.write(">%s\n%s\n" % (label, seq))
+                        okseqs += 1
```
wasade: this looks over indented? same with the else branch

deblur/workflow.py
```diff
@@ -382,40 +347,133 @@ def write_biom_table(table, biom_fp):
     biom_fp: string
         filepath to output BIOM table
     """
+    logger = logging.getLogger(__name__)
+    logger.debug('write_biom_table to file %s' % biom_fp)
     with biom_open(biom_fp, 'w') as f:
```
wasade: :( i hate this interface... my bad.

deblur/workflow.py
```diff
@@ -382,40 +347,133 @@ def write_biom_table(table, biom_fp):
...
+            name of fasta files to be added to the biom table
+            sampleid (file names without the file_end and path)
+    """
+    logger = logging.getLogger(__name__)
+    logger.debug('get_files_for_table input dir %s, '
+                 'file-ending %s' % (input_dir, file_end))
 
+    names = []
+    # scan the input directory entries
+    for cfile in listdir(input_dir):
```
wasade: so this does have a practical limit of 30k samples due to stress on the filesystem... but that can be revisited in a later PR. note, metadata checks (like `isfile` and `stat`) will get appreciably expensive and stressful to the filesystem under a large number of samples, but again, worth deferring to later

scripts/deblur
wasade: I think @josenavas was referring to the `__version__` variable

deblur/workflow.py
```diff
@@ -382,40 +346,133 @@ def write_biom_table(table, biom_fp):
...
+            name of fasta files to be added to the biom table
+            sampleid (file names without the file_end and path)
+    """
+    logger = logging.getLogger(__name__)
+    logger.debug('get_files_for_table input dir %s, '
+                 'file-ending %s' % (input_dir, file_end))
+
+    names = []
+    # scan the input directory entries
+    for cfile in listdir(input_dir):
```
josenavas: @amnona Sorry for not mentioning this earlier, I just remember about glob. I think this entire `for` loop can be simplified by using this code:

``` python
from os.path import join
from glob import glob
for cfile in glob(join(input_dir, "*%s" % file_end)):
    if not isfile(cfile):
        continue
    sample_id = cfile[:-len(file_end)]
    names.append((cfile, sampleid))
```

`glob` should already return all the paths that match the `"*.fasta.trim.derep.no_artifacts.msa.deblur.no_chimeras"` wildcard on the given directory. I tested it on a directory that I have and seems to do the same here, worth doublechecking though.
Also `cfile` will be the correct path already, so you will not need the join.

deblur/workflow.py
josenavas: Thanks @amnona ! This function looks way better now 😄

deblur/workflow.py
josenavas: Makes sense - thanks!

scripts/deblur
josenavas: Yup - @wasade is correct - sorry for not being clear - the version number is in my opinion a bit arbitrary although it is not used anywhere. However, that shouldn't be a blocker of this PR.

josenavas:
@amnona done with another pass of review - thanks for fixing everything - I only have one extra comment - do you think you can fix it soon? If not I can open an issue to fix later.
@wasade do you want to give another review or from your last one do you think it is ok?

wasade:
I'm :+1:

amnona:
fixed the last standing PR issues (now we even write the program version to the log :) )

josenavas:
thanks @amnona !!!

PR #35, opened 2016-07-12
fixed the parallel code, let's see if it passes travis :)

amnona:


deblur/deblurring.py
```diff
@@ -36,7 +36,10 @@ def get_sequences(input_seqs):
     seqs = [Sequence(label, seq) for label, seq in input_seqs]
 
     if len(seqs) == 0:
-        raise ValueError("No sequences found!")
+        logger = logging.getLogger(__name__)
+        logger.warn('No sequences found in fasta file!')
```
wasade: in a parallel context, shouldn't this logging message and the ones below denote the sample?

deblur/test/test_parallel_deblur.py
```diff
@@ -0,0 +1,121 @@
...
+        self.orig_s2_fp = join(self.test_data_dir, 'simset.s2.fasta')
+        self.orig_s3_fp = join(self.test_data_dir, 'simset.s3.fasta')
+
+        self.files_to_remove = []
+
+        logfilename = join(self.working_dir, "log.txt")
+        start_log(level=logging.DEBUG, filename=logfilename)
+
+    def tearDown(self):
+        pass
```
wasade: remove please

deblur/test/test_parallel_deblur.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        # index the 70% rep. set database
+        ref_fp = join(self.test_data_dir, '70_otus.fasta')
+        ref_db_fp = build_index_sortmerna(
+            ref_fp=(ref_fp,),
+            working_dir=self.working_dir)
+
+        trim_length = 100
+        params = ['deblur', 'workflow', '--seqs-fp', 'ignorethis', '--output-dir', self.working_dir, '--ref-fp',
+                  ref_fp, '-d', '1,0.06,0.02,0.02,0.01,0.005,0.005,0.005,0.001,0.001,0.001,0.0005', '-t', str(trim_length)]
```
wasade: possible to adjust these lines to fit in under 80 chars?

deblur/workflow.py
```diff
@@ -206,9 +206,14 @@ def remove_artifacts_seqs(seqs_fp,
                 # if * means no match
                 if line[1] == '*':
                     continue
-                # check if % identity and coverage are large enough
-                if (float(line[2]) >= sim_thresh*100) and (float(line[13]) >= coverage_thresh*100):
-                    aligned_seq_ids.add(line[0])
+                # check if % identity[2] and coverage[13] are large enough
+                # note e-value is [10]
+                if negate:
```
wasade: the addition of this statement seems quite important given the context. do you think this was the source of the parallel issue?

deblur/deblurring.py
amnona: Not trivial remembering the filename we're processing...
There is a ThreadID in the log messages.
If you know how to set a thread name (when it's created) we can set it to the file name and put also the thread name in the log messages

deblur/workflow.py
amnona: Don't think this is the source. It also worked without it.
Changed it following a talk with Jenya who suggested using the E value instead of the ident/coverage for identifying 16S based on greengenes

deblur/workflow.py
```diff
@@ -550,6 +555,10 @@ def launch_workflow(seqs_fp, working_dir, read_error, mean_error, error_dist,
     with open(output_deblur_fp, 'w') as f:
         seqs = deblur(parse_fasta(output_msa_fp), read_error, mean_error,
                       error_dist, indel_prob, indel_max)
+        if seqs is None:
+            warnings.warn('multiple sequence alignment file %s contains no sequences' % output_msa_fp, UserWarning)
```
amnona: You got me :) i disabled line length checking... think it makes the program less clear...
but fixed it anyway.

setup.py
```diff
@@ -46,8 +46,10 @@
       scripts=glob('scripts/*'),
       extras_require={'test': ["nose >= 0.10.1", "pep8"],
                       'doc': ["Sphinx >= 1.2.2", "sphinx-bootstrap-theme"]},
-      install_requires=['click', 'numpy >= 1.7',
+      install_requires=['click >= 6', 'numpy >= 1.7',
                         'scikit-bio >= 0.2.2, < 0.3.0',
                         'biom-format >= 2.1.3, < 2.2.0'],
+      dependency_links=[('https://github.com/pallets/click/archive/master.zip'
```
amnona: Removed the click 7 dependency - we don't use 'hidden' now (and don't want to depend on dev version). i think it should work with click 6 (didn't test 5)
Don't know what the 'dependecy_links' means or how to fix it...

deblur/deblurring.py
josenavas: Does the thread write the name of the file that is processing when it starts processing it? That may be enough to help to troubleshoot.

deblur/deblurring.py
```diff
@@ -36,7 +36,10 @@ def get_sequences(input_seqs):
...
+        return None
+#       raise ValueError("No sequences found!")
```
josenavas: Remove commented code.

deblur/parallel_deblur.py
```diff
@@ -0,0 +1,160 @@
...
+    script_name = "deblur"
+    script_subprogram = "workflow"
+    command = [script_name,
+               script_subprogram,
+               '--seqs-fp', input_fp,
+               '--is-worker-thread',
+               '--keep-tmp-files']
+    command.extend(params)
+
+    logger.debug('running command %s' % command)
```
josenavas: I _think_ this will write the name of the file. If the message has the thread ID, then we will be able to traceback which fasta file was empty.

deblur/parallel_deblur.py
```diff
@@ -0,0 +1,160 @@
...
+    return system_call(command)
+
+
+def system_call(cmd):
```
josenavas: There is already a `system_call` function in workflow.py: https://github.com/amnona/deblur/blob/23b55ba9671dd2f5f2050fe7764cbbfc71ab9e4a/deblur/workflow.py#L642

can you import it from there to avoid code duplication?

deblur/workflow.py
```diff
@@ -403,6 +419,12 @@ def create_otu_table(output_fp, deblurred_list):
     deblurred_list : list of string
         list of file names (including path) of all deblurred
         fasta files to add to the table
+    outputfasta_fp : str
```
josenavas: can you add `, optional` here and in the `minreads` parameter? Thanks!

setup.py
josenavas: @amnona you can completely delete the dependency links entry, since now we are depending on a release verion of click.

scripts/deblur
```diff
@@ -386,57 +432,85 @@ def build_biom_table(seqs_fp, output_biom_fp, file_type, log_level, log_file):
...
+            out_dir_split = join(output_dir, "split")
+            logger.info('splitting file %s to per sample fasta. '
+                        'output %s' % (seqs_fp, out_dir_split))
+            mkdir(out_dir_split)
+            with open(seqs_fp, 'U') as seqs_f:
+                split_sequence_file_on_sample_ids_to_files(
+                    seqs_f,
+                    out_dir_split)
+        input_file_list = glob('%s/*.fasta' % out_dir_split)
+#    input_file_list = listdir(out_dir_split)
```
josenavas: remove commented code

scripts/deblur
```diff
@@ -386,57 +432,85 @@ def build_biom_table(seqs_fp, output_biom_fp, file_type, log_level, log_file):
...
-              dir_okay=True), default='deblur.log',
+@click.option('--log-file', required=False,
+              type=click.Path(resolve_path=True, readable=True,
+                              exists=False, dir_okay=True),
+              default='deblur.log',
               show_default=True, help="log file name")
+@click.option('--jobs-to-start', '-O', required=False, type=int, default=1,
+              show_default=True,
+              help="Number of jobs to start (if to run in parallel)")
+@click.option('--is-worker-thread', required=False, type=bool, default=False,
```
josenavas: I think this option can mislead users. Usually, in multiprocessed programs, each thread has its own id (see [this](http://stackoverflow.com/a/10192611/3746629) SO answer to see how to get it in python multiprocessing) and then the thread with lowest ID (in this case 1) is the master thread, while the other ones are working threads. You can use that to remove this option and just manage that internally.

deblur/workflow.py
josenavas: Can you extend/modify the tests so we test both branches in this if? (that is, sometimes with negate=True, and others with negate=False)

josenavas:
Thanks @amnona done with the review! Do you know what did you change that made the tests to pass?

deblur/parallel_deblur.py
amnona: cool.
can follow the thread file from the threadid.
If someone knows how to set the thread name (we'll set it to the file name), i can print it in the log with each line, so it'll be nicer to follow the log file...

scripts/deblur
amnona: Cool idea, but maybe it's a bit too hackish? I'm not sure this is a guaranteed behavior in python, so may in some future version/some specific machine the main thread will not get id 1? or is it guaranteed?
What do you think is best to do?

amnona:
Not sure.
I didn't change the heart of the paralleliation (run_functor/mp.Pool)
however, i did rewrite all the test code, since it was not using the
correct deblur worflow functions. So maybe something there? I also rewrote
the way parallel deblur is run (integrated it into deblur workflow...). So
maybe one of these things did the magic?

PR #40, opened 2016-08-17
fix split problem for samples with _ in name

amnona:


josenavas:
Can you provide a bit more context of the problem? It is unclear to me how the changes in the code are fixing any issue with the title of the PR. Also, this kind of changes require a specific test case to make sure that future changes do not incur in the same issue and to exemplify better the origin of the issue.

Thanks!

amnona:
The problem is when samples have names with underline (e.g. sample_1_is_me,sample_2_is_me)
the old version would split and take the right most part (i.e. 'sample') instead of the full name (each read is labelled sample_1_is_me_1, sample_1_is_me_2,sample_1_is_me_3,...)
instead we take all the sample name ( using rsplit('_',1) )
will try to add a test

josenavas:
Thanks, what about the addition of `split(' ')[0]`?

deblur/workflow.py
```diff
@@ -324,6 +324,29 @@ def remove_chimeras_denovo_from_seqs(seqs_fp, working_dir, threads=1):
...
+        the fasta file read name
+
+    Returns
+    -------
+    sampleid : str
+        the sample id
+    """
+
+    # get the sampleid_readid field
+    sampleread=readid.split(' ')[0]
```
josenavas: Add spaces around the '=' char - and same below.

deblur/test/test_workflow.py
```diff
@@ -572,6 +573,22 @@ def get_seqs_act_split_sequence_on_sample_ids(self, output_dir):
...
 
+    def test_sample_id_from_read_id(self):
+        """Test the fasta readid to sample id
+        used in split_sequence_file_on_sample_ids_to_files
+        """
+        self.assertEqual(sample_id_from_read_id("Samp1_0 M04771:27:000000000"
+                                                "-ARFWH:1:1101:18081:1897 1:"
+                                                "N:0:0 orig_bc=CGTTAAGTCAGC n"
+                                                "ew_bc=CGTTAAGTCAGC bc_diffs="
+                                                "0"),"Samp1")
```
josenavas: Add a space after the ',', same below

josenavas:
@amnona 2 really small comments - after that it should be good to go

wasade:
Any luck removing the out-of-context commits?

amnona:
Talked to Yoshiki. Now I understand how to do it next time :)
To fix the current situation is more complicated...need Yoshiki to help, so
will only be able to do it Monday.
Alternatively If  someone can go over the multiple commits in this PR and
merge it, it'll be great (it's not so many changes. Just looks scary
because one commit added the greengenes 88 fasta file - so it shows as a
lot of lines added...)

scripts/deblur
```diff
@@ -77,12 +87,33 @@ def deblur_seqs(seqs_fp, read_error, mean_error, error_dist, indel_prob,
...
+            error_dist = create_error_profile(read_error)
+    else:
+        if read_error is not None:
+            # if we get both read_error and error_dist it's an error
+            # since they both set the same variable
+            logger.error('Cannot use both --error_dist and --read_error')
+            raise ValueError("Cannot use both --error_dist and --read_error")
+
+    logger.info('error_dist is : %s' % error_dist)
+    error_dist = list(map(float, error_dist.split(',')))
```
wasade: This call will raise an exception when `error_dist is None` and `read_error is not None` as the return type from `create_error_profile` is a `list` of `float` and not a `str`.

scripts/deblur
```diff
@@ -34,6 +34,9 @@ from deblur.workflow import (launch_workflow, trim_seqs, dereplicate_seqs,
 
 __version__ = "0.9.2.1"
 
+default_error_dist = '1,0.06,0.02,0.02,0.01,0.005,0.005' \
```
wasade: What mechanism ensures this default is the same default used by `deblur`?

deblur/deblurring.py
```diff
@@ -55,7 +55,34 @@ def get_sequences(input_seqs):
...
+    """
+    error_profile = [1.0, read_error, np.max([0.01, np.power(read_error, 2)]),
+                     np.max([0.01, np.power(read_error, 3)]),
+                     0.01, 0.005, 0.005, 0.005, 0.005, 0.005, 0.005, 0.001,
+                     0.001, 0.001, 0.001]
+    return error_profile
+
+
+def deblur(input_seqs, mean_error=0.005,
+           error_dist=[1, 0.05, 0.000005, 0.000005, 0.000005, 0.000005,
```
wasade: Why is this default `error_dist` different from the "default" generated in the `deblur_seqs` when `error_dist is None` and `read_error is None`?

deblur/deblurring.py
```diff
@@ -105,13 +128,7 @@ def deblur(input_seqs, read_error=0.05, mean_error=0.005, error_dist=None,
 
     # if error_list not supplied, use the default (22 mock mixture setup)
```
wasade: this comment is not accurate

scripts/deblur
```diff
@@ -391,21 +422,23 @@ def build_biom_table(seqs_fp, output_biom_fp, min_reads, file_type, log_level, l
 @click.option('--overwrite', '-w', required=False, type=bool, default=False,
               is_flag=True,
               show_default=True, help="Overwrite output directory if exists.")
-@click.option('--read-error', '-e', required=False, type=float, default=0.05,
+@click.option('--read-error', required=False, type=float, default=None,
               show_default=True,
               help="Maximal bound on 1 nucleotide read error rate."
-                   "Required if --eror-dist is not supplied")
+                   "If supplied, overrides the --eror-dist!")
```
wasade: `--error-dist`

Also, please omit the `!` as it is not polite to yell at the user

scripts/deblur
```diff
@@ -391,21 +422,23 @@ def build_biom_table(seqs_fp, output_biom_fp, min_reads, file_type, log_level, l
...
-@click.option('--error-dist', '-d', required=False, type=str, default=None,
-              show_default=True,
+@click.option('--error-dist', '-d', required=False, type=str,
+              default=None,
+              show_default=False,
               help="A comma separated list of error probabilities for each "
                    "hamming distance. The length of the list determines the "
                    "number of hamming distances taken into account. "
-                   "This replaces of the --read-error value")
+                   "Ignored if --read-error is supplied.\n"
```
wasade: This is inaccurate: it is not ignored, an exception is raised. Suggest simply stating that this option is exclusive w.r.t. `--read-error`

scripts/deblur
```diff
@@ -524,7 +575,7 @@ def workflow(seqs_fp, output_dir, ref_fp, ref_db_fp, overwrite,
             if isfile(input_fp):
                 deblurred_file_name = launch_workflow(
                     seqs_fp=input_fp, working_dir=working_dir,
-                    read_error=read_error, mean_error=mean_error,
```
wasade: `read_error` is no longer used?

wasade: nm, i see now

deblur/workflow.py
```diff
@@ -40,11 +40,25 @@ def trim_seqs(input_seqs, trim_len):
...
 
     for label, seq in input_seqs:
+        totseqs += 1
         if len(seq) >= trim_len:
+            okseqs += 1
             yield label, seq[:trim_len]
 
+    if okseqs < 0.01*totseqs:
+        logger = logging.getLogger(__name__)
+        logger.warn('Vast majority of sequences are shorter than the '
```
wasade: Why not report the percentage of sequences?

wasade:
What is the context for these other commits? There's a totally new function, `create_error_profile` as well as parameter and method changes to `deblur`.

deblur/deblurring.py
```diff
@@ -55,7 +55,34 @@ def get_sequences(input_seqs):
     return seqs
 
 
-def deblur(input_seqs, read_error=0.05, mean_error=0.005, error_dist=None,
+def create_error_profile(read_error):
+    """Create a theoretical read error profile given
+    a maxima error rate
```
mortonjt: Can you elaborate on what exactly the error profile is?  

For instance, how was the length of the error profile chosen?

scripts/deblur
amnona: Nice catch!
Would you recommend to have create_error_profile return a string?
or do the error_dist = list(map(float, error_dist.split(','))) inside the if/else
or maybe have create_error_profile get as input both error_dist and read_error and return the float list of read errors?

scripts/deblur
amnona: We want to have a default error profile (from our experience with illumina and mock mixture analysis). However, we want to allow the user 2 alternative ways to supply his own error profile:
1. write it as a comma separated list for each hamming distance
2. supply a maximal hamming 1 read error (read_error) and create the hamming 2,3 etc. profile from it (less good since the assumption you can predict hamming 2 error prob. from hamming 1 is incorrrect).
* But the user cannot supply both together.

Since if we supply the default error profile as a default value to error_dist, we won't be able to sense if the user supplied both error_dist and read_error and warn about it (which seems to me some users will make this mistake), we instead have the default error profile as default_error_dist, and if error_dist is None, we do error_dist = default_error_dist

But other ideas will be welcome :)

wasade: What I'm concerned about is that this "default" is specified in multiple places, and there does not appear to be any mechanism to protect against a future developer making a change in one place and not another

scripts/deblur
wasade: The default `error_dist` is imported and is in its native type already. `create_error_profile` should return its native type. Any formatting for strings, as is necessary for dumping details on the command line, should be done within the `help=` text formatting for the relevant option

deblur/deblurring.py
amnona: We want to be able to write the default error profile in the deblur workflow --help. If we would just use the defaults in this function, we won't know what to print...
Maybe best to have a function that returns the default error profile, and have deblur() just call it if error_dist is None?
Anyway, this default inside deblur() is not supposed to be used when running via the command line

wasade: why is this default not appropriate to use when running from the command line?

amnona:
commits:
- fix split problem for samples with _ in name (81bd726),
- fix split problem for samples with _ in name (4ab87d7),
- samplename from readid as a function+test (6159a64),
- pep8 fun (4cd1458),
- pep8 fun again (987eeff)
Are all iterations of fixing the split_sequence_file_on_sample_ids_to_files() when the sample name contains '_'


commit:
-  the positive filtering greengenes fasta file (699dced)
Adds the greengenes 88% rep_set fasta file to the repository (for positive filtering)

commit:
- remove the --delim option (not needed and not used) (5366747)
removes the --delim option for the split_sequence_file_on_sample_ids_to_files() since it was not used and complicates life for the end user

commit:
- add warning for too short trim length (d6fd2d9)
Adds a warning when the trim length seems to be too long - so the user will know why deblur failed.

commit:
- add default illumina error profile (e4333b6)
Inserts the default error profile when the user doesn't supply one (and also moved the read_error based error profile generation to a separate function inside deblurring.py)

deblur/deblurring.py
amnona: it is a comma separated list of upper bound on error probability per hamming distance.
let's denote p(d) the upper bound on the probability to get d changes in the sequence (hamming distance). the error probability is [p(0),p(1),p(2),p(3),...p(n)].
while ideally p(2) should be p(1)^2, it is not.
n is determined by when the error probability gets so small it's practically 0 (we use n=11, but p(10) is set to very small so not critical - p is decreasing fast).
the error profile is based on phix reads, mock mixtures and samples with known bacteria.

deblur/deblurring.py
amnona: fixed

scripts/deblur
amnona: ok, won't shout at the user :)

scripts/deblur
amnona: fixed

deblur/workflow.py
amnona: fixed

mortonjt:
@amnona and I are currently pair-programming this.  Looks like the tests are passing.  So right now, we are just cleaning up the code and adding more tests to bump up the coverage.

amnona:
Ok, we have added more unit test coverage (thanks @mortonjt !) and a few more log messages.
should we add it to this PR (and make it more complicated) or as a separate PR?

wasade:
@amnona, coverage is showing a decrease on this PR, are all of the changes in?

amnona:
there is a separate branch with the additional coverage unit tests. didn't merge in since don't want to load too much on this PR

wasade:
is this good to review?

amnona:
👍 ready for review :)

deblur/test/test_deblurring.py
```diff
@@ -87,6 +99,41 @@ def test_deblur(self):
 
         self.assertEqual(obs, exp)
 
+    def test_deblur_indel(self):
+        """Test if also removes indel sequences
+        """
+        seqs_f = StringIO(TEST_SEQS_2)
+
+        # add the MSA for the indel
+        seqs = parse_fasta(seqs_f)
```
wasade: There's going to be conflicts depending on what gets merged first...

deblur/deblurring.py
```diff
@@ -55,7 +65,31 @@ def get_sequences(input_seqs):
...
+        the upper bound on the hamming one read error
+
+    Results
+    -------
+    error_profile : list of float
+        the predicted theoretical error profile
+        (note for hamming>1 we take the max of
+        read_error^2 and 0.01 since closer to reality)
+    """
+    error_profile = [1.0, read_error, np.max([0.01, np.power(read_error, 2)]),
```
amnona: following discussion with @mortonjt , we intend to remove the --read-error option and get rid of the create_error_profile() function.
It has no real use (since read errors are not independent - as is assumed in the create_error_profile() function), and it can confuse the users.
This option was used for comparisons to the illumina error profile (to see how different it is from the simple assumptions), and i don't think it should be used anymore.
Just didn't want to complicate too many things on this PR 
so will remove it in a following PR

deblur/test/test_deblurring.py
```diff
@@ -87,6 +99,41 @@ def test_deblur(self):
...
+        newseqs = []
+        for chead, cseq in seqs:
+            tseq = cseq[:10] + '-' + cseq[10:]
+            newseqs.append((chead, tseq))
+        # now add a sequence with an A insertion
+        tseq = cseq[:10] + 'A' + cseq[10:-1]+'-'
+        newseqs.append((chead, tseq))
+
+        # cheads, seqs = zip(*newseqs)
```
wasade: can these commented out lines be deleted please?

deblur/test/test_workflow.py
```diff
@@ -439,23 +450,36 @@ def run_workflow_try(self, simfilename, origfilename, ref_fp, ref_db_fp):
...
+            number of threads to use (default=1)
+        trim_length : int
+            length of sequences to trim to (default=100)
         """
         seqs_fp = simfilename
         output_fp = self.working_dir
-        read_error = 0.05
         mean_error = 0.005
-        error_dist = None
+        error_dist = [1, 0.05, 0.000005, 0.000005, 0.000005, 0.000005,
```
wasade: is this a different default error profile?

scripts/deblur
amnona: fixed using error_dist_from_str() and click callback option

deblur/test/test_workflow.py
```diff
@@ -522,6 +548,51 @@ def test_create_otu_table(self):
...
+        outfile2 = join(self.working_dir, 'testmerge2.biom')
+        create_otu_table(outfile2, [(m1, 'testmerge'), (m2, 'testmerge2')],
+                         minreads=minreads)
+
+        table2 = load_table(outfile2)
+        table2ids = table2.ids(axis='observation')
+        tablesum = table.sum(axis='observation')
+        for idx, cid in enumerate(table.ids(axis='observation')):
+            if tablesum[idx] >= minreads:
+                self.assertTrue(cid in table2ids)
```
wasade: I believe there is a `self.assertContains` or something along those lines

scripts/deblur
```diff
@@ -45,16 +61,24 @@ def deblur_cmds():
 @click.argument('seqs_fp', required=True,
                 type=click.Path(resolve_path=True, readable=True, exists=True,
                                 file_okay=True))
-@click.option('--read-error', '-e', required=False, type=float, default=0.05,
-              help="Maximal bound on 1 nucleotide read error rate")
+@click.option('--read-error', required=False, type=float, default=None,
+              show_default=True,
+              help="Maximal bound on 1 nucleotide read error rate."
+                   "If supplied, overrides the default --eror-dist")
```
wasade: `--eror-dist` -> `--error-dist`

scripts/deblur
wasade: this call to `list(map(...))` shouldn't be necessary anymore, right?

scripts/deblur
```diff
@@ -391,21 +436,24 @@ def build_biom_table(seqs_fp, output_biom_fp, min_reads, file_type, log_level, l
 @click.option('--overwrite', '-w', required=False, type=bool, default=False,
               is_flag=True,
               show_default=True, help="Overwrite output directory if exists.")
-@click.option('--read-error', '-e', required=False, type=float, default=0.05,
+@click.option('--read-error', required=False, type=float, default=None,
               show_default=True,
               help="Maximal bound on 1 nucleotide read error rate."
-                   "Required if --eror-dist is not supplied")
+                   "If supplied, overrides the default --eror-dist")
```
wasade: same comment on `--eror-dist`

deblur/test/test_deblurring.py
amnona: fixed

wasade:
Few comments. I'm still confused by the default error profile as it looks like its defined in multiple places, and it doesn't look like its possible to derive it from the `create_error_profile` function

deblur/deblurring.py
mortonjt: Since this is being phased out, we can delete this right?

deblur/test/test_deblurring.py
mortonjt: yeah ... I noticed that when building the conda recipe.

deblur/deblurring.py
mortonjt: I think it is ok to ditch `create_error_profile()` in this PR, since that is going to be kicked out anyways.

deblur/test/test_workflow.py
mortonjt: I've never seen an assertContains.

But I know that there is an [assertIn](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn)

mortonjt: This can be changed to
```
self.assertIn(cid, table2ids)
```

deblur/test/test_workflow.py
```diff
@@ -522,6 +547,51 @@ def test_create_otu_table(self):
...
+            else:
+                self.assertTrue(cid not in table2ids)
```
mortonjt: This can be changed to
```
self.assertNotIn(cid, table2ids)
```

mortonjt:
Looks good 👍

PR #59, opened 2016-08-23
MAINT: removed HAVE_H5PY check

### LINKED ISSUE #51: saving to json (when h5py isn't present) doesn't actually work
Strongly recommend removing biom 1.x format support and requiring h5py.

wasade:
Fixes #51 

Note, `h5py` was already listed as a requirement in `setup.py` so the `HAVE_H5PY` check would always evaluate `True`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
   # Update conda itself
   - conda update --yes conda
 install:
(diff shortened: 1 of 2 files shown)


mortonjt:
👍

amnona:
byebye json

PR #62, opened 2016-08-23
Default fasta files for positive and negative filtering

amnona:
- Added the greengenes 88 rep set for positive filtering (keeping only sequences close enough to it)
- Added the artifacts.fa (which has phix and the adapter sequence) for negative filtering (remove sequences matching to this file)
into  the support_files directory.
- support_files also has an __init__.py which sets the paths to these files in pos_db and neg_db variables
- added the deblur.support_files package to the setup.py and the files into the package_files variable
- set the default values for the --ref-fp parameter according to filtering method (-n flag)
- fix a small error in the cli for default read error profile


scripts/deblur
```diff
@@ -40,6 +41,8 @@ def error_dist_from_str(ctx, param, value):
     """
     if value is None:
         return value
+    if type(value) != str:
```
mortonjt: Can replace this with

`not isinstance(value, str)`

To tidy this up

deblur/support_files/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+# Distributed under the terms of the BSD 3-clause License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+import os
+
+pos_db = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                      '88_otus.fasta')
+neg_db = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                      'artifacts.fa')
```
wasade: can you add:

```python
__all__ = ['pos_db', 'neg_db'] 
```

please?

deblur/test/test_support_files.py
```diff
@@ -0,0 +1,36 @@
...
+from deblur.support_files import pos_db, neg_db
+
+import os.path
+
+
+class supportFilesTests(TestCase):
+    """Test the supporting data files
+    """
+
+    def setUp(self):
```
wasade: these can be omitted since they're noop

wasade:
2 minor comments, :+1: otherwise

wasade:
:+1:, will merge once CI is complete

PR #67, opened 2016-08-26
Update README.md

amnona:
if someone can translate it into real English :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,64 @@ Deblur
 [![Build Status](https://travis-ci.org/biocore/deblur.png?branch=master)](https://travis-ci.org/biocore/deblur)
 [![Coverage Status](https://coveralls.io/repos/github/biocore/deblur/badge.svg?branch=master)](https://coveralls.io/github/biocore/deblur?branch=master)
 
-Deblur is a greedy deconvolution algorithm based on known read error profiles.
+Deblur is a greedy deconvolution algorithm based on Illumina Miseq/Hiseq error profiles.
+
+Install
+=======
+- Deblur requires Python 3.5. If Python 3.5 is not installed, you can create a conda environment for deblur using:
+
+```
+conda create -n debenv python=3
+```
+
+and activate it using:
+
+```
+source activate debenv
+```
+
+- Install Deblur dependencies:
+```
+conda install -c biocore VSEARCH MAFFT SortMeRNA numpy
+```
+
+- Install Deblur:
+```
+pip install git+git://github.com/biocore/deblur
+```
+
+Example usage
+=============
+
+The input to deblur workflow is a directory of fasta files (1 per sample) or a demultiplexed FASTA or FASTQ file. The output is a biom table with sequences as the OTU ids (final.biom in the output directory).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,30 +4,69 @@ Deblur
...
+
+- Install Deblur dependencies:
+```
 conda install -c biocore VSEARCH MAFFT SortMeRNA numpy
-python setup.py install
 ```
 
+- Install Deblur:
+```
+pip install git+git://github.com/biocore/deblur
```
wasade: No, we need this to be either:

`pip install deblur`

or 

`conda install deblur` (possibly with `-c biocore`)

The reason is that this install method is not against a specific version but against the repository itself

README.md
```diff
@@ -4,30 +4,69 @@ Deblur
...
-algorithm is designed to work on individual samples.
+The input to deblur workflow is a directory of fasta files (1 per sample) or a demultiplexed FASTA or FASTQ file. The output is a biom table with sequences as the OTU ids (final.biom in the output directory).
+
+The simple use case just specifies the input fasta file (or directory) and output directory name:
+
+```
+deblur workflow --seqs-fp all_samples.fna --output-dir output
+```
+
+If starting from a barcode and read file, you can first use the qiime split_libraries_fastq.py command (we recommend using -q 19 to remove low quality reads):
```
wasade: Should the default for split libraries in qiime be changed to `-q 19`, or is this specific to deblur?

README.md
```diff
@@ -4,30 +4,69 @@ Deblur
...
+
+```
+split_libraries_fastq.py -i XXX_R1_001.fastq -m map.txt -o split -b XXXX_I1_001.fastq -q 19
+```
+
+and use the split/seqs.fna as the input to the deblur workflow.
+
+Positive and Negative Filtering
+===============================
+By default, deblur uses positive filtering, keeping only 16S sequences (based on homology to greengenes 88% representative set). for example:
```
wasade: Greengenes

wasade: for -> For

README.md
```diff
@@ -4,30 +4,69 @@ Deblur
...
 
 ```
-deblur workflow --seqs-fp all_samples.fna --ref-fp db.fna --output-dir output
+deblur workflow --seqs-fp all_samples.fna --output-dir output
+```
+
+Negative filtering can be selected using the '-n' flag. This causes deblurring to keep all sequences except known artifact sequences (i.e. PhiX and Adapter sequences), so other non-16S sequences are retained. For example:
```
wasade: "except known" -> "except for known"

wasade: would it be possible to add a hyperlink to the negative filtering fasta file used by deblur? something like, "the specific sequences used for filtering can be found here"

wasade:
I think this is really good. Just a few comments

README.md
amnona: i think in general it is a good approach to life, but changing the default in qiime is a bigger question - probably best to hear what other people have to say (for deblurring it's not critical - but it think may be slightly better)

amnona:
closing since covered in PR #68

PR #68, opened 2016-08-26
add hyperlinks and small fixes

amnona:
fixed all comments from PR #67 
also added more links and important options

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,75 @@ Deblur
 [![Build Status](https://travis-ci.org/biocore/deblur.png?branch=master)](https://travis-ci.org/biocore/deblur)
 [![Coverage Status](https://coveralls.io/repos/github/biocore/deblur/badge.svg?branch=master)](https://coveralls.io/github/biocore/deblur?branch=master)
 
-Deblur is a greedy deconvolution algorithm based on known read error profiles.
+Deblur is a greedy deconvolution algorithm based on Illumina Miseq/Hiseq error profiles.
+
+Install
+=======
+- Deblur requires Python 3.5. If Python 3.5 is not installed, you can create a conda environment for deblur using:
+
+```
+conda create -n debenv python=3
+```
+
+and activate it using:
+
+```
+source activate debenv
+```
+
+- Install Deblur dependencies:
+```
+conda install -c biocore VSEARCH MAFFT SortMeRNA numpy
+```
+
+- Install Deblur:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,30 +4,79 @@ Deblur
...
 ```
-deblur workflow --seqs-fp all_samples.fna --ref-fp db.fna --output-dir output
+split_libraries_fastq.py -i XXX_R1_001.fastq -m map.txt -o split -b XXXX_I1_001.fastq -q 19
 ```
+
+and use the split/seqs.fna as the input to the deblur workflow.
+
+Important options
+=================
+- the sequence read length can be specified by the ```-t NNN``` flag, where NNN denotes the length all sequences will be trimmed to (default=150). Note that all reads shorter than this length will be discarded.
```
mortonjt: the - The

mortonjt:
👍

README.md
```diff
@@ -4,30 +4,79 @@ Deblur
...
+and activate it using:
 
 ```
+source activate debenv
+```
+
+- Install Deblur dependencies:
+```
 conda install -c biocore VSEARCH MAFFT SortMeRNA numpy
-python setup.py install
```
mortonjt: Can you add
```
conda install -c bioconda biom-format 
```

README.md
```diff
@@ -4,32 +4,72 @@ Deblur
...
-=====================
-
-Some of the code in the package deblur has been derived from [QIIME](http://qiime.org).
-The contributors to these specific QIIME modules have granted permission
-for this porting to take place and put under the BSD license.
+Deblur is a greedy deconvolution algorithm based on Illumina Miseq/Hiseq error profiles.
 
 Install
 =======
+- Deblur requires Python 3.5. If Python 3.5 is not installed, you can create a conda environment for deblur using:
```
wasade: The below doesn't create the environment but installs packages into the current environment, so the sentence is inaccurate

wasade:
Would it be possible to add a note about how to ask for help? Not sure if we should setup a google group for this or centralize on the repo

mortonjt:
I think it is ok to refer to people to the issue tracker for now.  What do you think?

mortonjt:
I think this is good to go

README.md
amnona: oops, disappeared in the conflict merge...
will fix.

PR #76, opened 2016-08-29
bugfix: 0 read fasta raised an error instead of continue

amnona:
following move to scikit-bio, when encountering a 0 read file (after msa), deblur raised an error instead of continuing with a warning.
fixed to write a warning and continue (so samples with 1 read will not stop the deblurring)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,20 @@
 # deblur changelog
 
+
+## Version 0.1.1-dev 
+### Features
+
+### Backward-incompatible changes [stable]
+
+### Performance enhancements
+
+### Bug fixes
+Fix total pipeline failure when a sample contains 1 read after dereplcation+singleton removal. Now issues a warning and continues [#75](https://github.com/biocore/deblur/issues/75)
+
+### Miscellaneous
+
 ## Version 0.1.0-dev 
+First release
 
 ### Features
 
--- a/deblur/test/test_workflow.py
+++ b/deblur/test/test_workflow.py
@@ -13,7 +13,6 @@ from os import listdir, remove
 from types import GeneratorType
 from os.path import join, isfile, abspath, dirname, splitext
 
-import skbio
 from skbio import DNA
 
 from biom import load_table
@@ -91,8 +90,14 @@ class workflowTests(TestCase):
         self.assertTrue(obs[0][1].startswith(first_few_nucs))
 
     def test_sequence_generator_invalid_format(self):
-        with self.assertRaises(skbio.io.FormatIdentificationWarning):
-            next(sequence_generator(join(self.test_data_dir, 'readme.txt')))
+        allres = []
+        input_fp = join(self.test_data_dir, 'readme.txt')
+        msg = "input file %s does not appear to be FASTA or FASTQ" % input_fp
+        with self.assertWarns(UserWarning) as W:
+            for res in sequence_generator(input_fp):
+                allres.append(res)
+        self.assertEqual(len(allres), 0)
+        self.assertEqual(str(W.warning), msg)
 
     def test_trim_seqs(self):
         seqs = [("seq1", "tagggcaagactccatggtatga"),
--- a/deblur/workflow.py
+++ b/deblur/workflow.py
@@ -69,8 +69,13 @@ def sequence_generator(input_fp):
         # http://scikit-bio.org/docs/latest/generated/skbio.io.format.fastq.html#format-parameters
         kw['variant'] = 'illumina1.8'
     else:
-        raise skbio.io.FormatIdentificationWarning("input_fp does not appear "
-                                                   "to be FASTA or FASTQ.")
+        # usually happens when the fasta file is empty
+        # so need to return no sequences (and warn)
+        logger = logging.getLogger(__name__)
+        msg = "input file %s does not appear to be FASTA or FASTQ" % input_fp
+        logger.warn(msg)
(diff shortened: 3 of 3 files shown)


mortonjt:
Looks like there is a failing test

amnona:
oops. forgot to fix the unittest.
should work now (test warn instead of raise)

amnona:
ok, now passes all checks :)

amnona:
Also, can we push this change to the pip install (and bump up the version :) )

mortonjt:
I'd hold off on the pip install - it gets confusing if there are too many releases.  Looks like the change is rather small - so I'd suggest waiting until a whole bunch of patches are made to shoot off another release.

And the dev version is still pip installable via

```
pip install git+https://github.com/biocore/deblur.git
```

So it won't be a problem pip installing the dev version.  This can be added to the readme if necessary.

deblur/test/test_workflow.py
```diff
@@ -91,8 +90,11 @@ def test_sequence_generator_fastq(self):
         self.assertTrue(obs[0][1].startswith(first_few_nucs))
 
     def test_sequence_generator_invalid_format(self):
-        with self.assertRaises(skbio.io.FormatIdentificationWarning):
-            next(sequence_generator(join(self.test_data_dir, 'readme.txt')))
+        allres = []
+        with self.assertWarns(UserWarning):
```
mortonjt: Can you add a warning message here?   It can be something like

```
msg = "input file readme.txt does not appear to be FASTA or FASTQ"
with self.assertWarns(UserWarning, msg):
...
```

deblur/workflow.py
```diff
@@ -69,8 +69,12 @@ def sequence_generator(input_fp):
         # http://scikit-bio.org/docs/latest/generated/skbio.io.format.fastq.html#format-parameters
         kw['variant'] = 'illumina1.8'
     else:
-        raise skbio.io.FormatIdentificationWarning("input_fp does not appear "
-                                                   "to be FASTA or FASTQ.")
+        # usually happens when the fasta file is empty
+        # so need to return no sequences (and warn)
+        logger = logging.getLogger(__name__)
+        logger.warn("input file %s does not appear to be FASTA or FASTQ" % input_fp)
+        warnings.warn("input file %s does not appear to be FASTA or FASTQ" % input_fp, UserWarning)
```
mortonjt: Can we simply these two lines to 
```python
msg = "input file %s does not appear to be FASTA or FASTQ" % input_fp
logger.warn(msg)
warnings.warn(msg, UserWarning)
```

mortonjt:
good to merge once comments are addressed

mortonjt:
Forgot to mention.

Now we are post alpha release, it is advantageous to start squashing commits.  It'll simplify the process of debugging using tools such as [`git bisect`](https://git-scm.com/docs/git-bisect).

You can do this by
```
git rebase -i 675f7fb
# rebase all commits except the first commit with `squash`
git push -f
```

Note that this is optional, but it will clean up the commit history.

wasade:
Can a change log mention be added please?

deblur/test/test_workflow.py
amnona: don't think we need to test the specific warning msg. text.

amnona:
Fixed (except specific warn msg text testing in unit_test)
tried to squash (not sure it worked)

amnona:
Since this is a BUGFIX, i think we should immediately update the pip (since the version is already out). Don't see any advantage in waiting?

wasade:
@amnona, do you perceive this as impacting most users or just a small percent? If the latter, then we might want to hold off another few days to see if any other bugs pop up which we should fix immediately.

ChangeLog.md
```diff
@@ -11,3 +12,16 @@
 ### Bug fixes
 
 ### Miscellaneous
+
+## Version 0.1.1-dev
```
wasade: can the most recent notes be put at the top of the file so the user sees the most recent changes first? normally the flow is:

```text
## version 1.2

### Features

* this really awesome thing #123

### Bug Fixes

* this really nasty thing happened where blah blah, see #20 for more detail

## Version 1.1

...
```

ChangeLog.md
```diff
@@ -11,3 +12,16 @@
...
+## Version 0.1.1-dev 
+
+### Features
+
+### Backward-incompatible changes [stable]
+
+### Performance enhancements
+
+### Bug fixes
+Fix total pipeline failure when a sample contains 1 read after dereplcation+singleton removal. Now issues a warning and continues
```
wasade: Can you please associate the specific PR? The scikit-bio ChangeLog is a great reference, and it makes it trivial for a user to look up the specific changes if necessary

deblur/test/test_workflow.py
mortonjt: This is advantageous to do. There could be a scenario where there are
multiple warnings raised by the function,  but the wrong warnings are
raised.  It's better safe than sorry.

amnona:
ready for merge :)

ChangeLog.md
josenavas: @amnona this comment has not been addressed

deblur/test/test_workflow.py
josenavas: Agree with @mortonjt if you're testing that the warning is raised, you need to make sure that the warning that you are catching here is the one that you're raising

PR #86, opened 2016-10-10
Update README.md

mortonjt:
Given that the biocore versions are broken, this will refer to bioconda instead


wasade:
I thought the bioconda ones did not work on Centos 6?

PR #87, opened 2016-10-11
add post-deblur positive filtering for negative mode

amnona:
after creating the negative mode (only artifacts) biom table and fasta file, run positive filtering and create two new biom tables - one with only 16s, one with the artifacts


deblur/workflow.py
```diff
@@ -207,6 +208,65 @@ def build_index_sortmerna(ref_fp, working_dir):
...
+        if cseq not in good_seqs:
+            good_seqs.add(cseq)
+    logger.debug('loaded %d sequences from cleaned biom table'
+                 ' fasta file' % len(good_seqs))
+
+    logger.debug('loading biom table %s' % table_filename)
+    table = load_table(table_filename)
+
+    # filter and save the artifact biom table
+    artifact_table = table.filter(list(good_seqs),
```
wasade: biom `filter` will not remove any IDs on the opposite axis which are entirely zero after an operation. do you think it would make sense to (in this case) remove samples which contain no data from the table? the reason this is important is any downstream consumer method which leverages `len(table.ids())` will be wrong

deblur/workflow.py
```diff
@@ -207,6 +208,65 @@ def build_index_sortmerna(ref_fp, working_dir):
...
+                                  axis='observation', inplace=False,
+                                  invert=True)
+    output_artifact_fp = join(biom_table_dir, 'final.only-non16s.biom')
+    write_biom_table(artifact_table, output_artifact_fp)
+    logger.info('wrote artifact only filtered biom table to %s'
+                % output_artifact_fp)
+
+    # filter and save the only 16s biom table
+    table.filter(list(good_seqs), axis='observation')
```
wasade: same comment here

deblur/workflow.py
amnona: in general deblur does not do any sample filtering (the old version used to keep also samples with 0 reads... not sure about this version). Anyway, a first step in any downstream analysis should be removing samples with low reads/rarifying.
We can incorporate this into deblur, but i don't think it's the best approach (and where do we stop - filter samples with 0 reads only? or filter also samples with 1 read...)
what do you think?

amnona: personally i prefer having the 0 read samples in the biom table, since this way i can know from the biom table how many samples were problematic. and since we don't gain anything be removing the 0 reads (since we still need to remove the low count samples), i think it's better the biom table contain all the information?

mortonjt: I agree with @amnona here.  Its pretty easy to filter out samples after the biom table is generated.  But it'll just become a little tedious if the user has to backtrack to the deblur log files to figure out which samples actually failed.

wasade:
All of the OTU picking methods in QIIME produce BIOM tables such that samples without viable sequence are not represented. It's not hard to determine what samples do not have sequence as you can refer back to the mapping file. It is very likely the case that there are downstream consumers of BIOM which make the implicit assumption that an ID which exists in the table actually has data; this an edge case which we are deviating from which could lead to unexpected consequences later.

antgonza:
Just adding to @wasade's point, in the past we have found "issues" when dealing with sample/OTUs with zero count, for example: https://github.com/biocore/qiime/issues/642. IMOO better safe than sorry.

mortonjt:
@amnona what do you think about creating a separate file to store the filtered out samples/sequences?  @antgonza has a point - it isn't intuitive what sort of exact filtering should be done.  And this could lead to other issues (i.e. divide by zero errors).  It could simplify downstream analyses if this information was moved out of the biom table into a separate file from the deblur output.

deblur/workflow.py
mortonjt: Looks like this has been addressed

scripts/deblur
```diff
@@ -411,21 +412,38 @@ def build_biom_table(seqs_fp, output_biom_fp, min_reads, file_type, log_level,
...
-                   (pos_db, neg_db))
-@click.option('--ref-db-fp', required=False, multiple=True,
+                   "default is: %s\n" % pos_db)
+@click.option('--pos-ref-db-fp', required=False, multiple=True,
               type=click.Path(resolve_path=True, readable=True, exists=False,
                               file_okay=True),
-              help="Keep all sequences aligning to this indexed "
-                   "database. For multiple databases, the order "
+              help="The indexed  version of the positive filtering database. "
+                   "If not supllied, deblur will index the database."
```
mortonjt: `supllied` -> `supplied`

scripts/deblur
```diff
@@ -411,21 +412,38 @@ def build_biom_table(seqs_fp, output_biom_fp, min_reads, file_type, log_level,
...
+              help="Negative (artifacts) filtering database. "
+                   "Keep all sequences aligning to this FASTA file "
+                   "(for multiple databases, use "
+                   "--ref-fp db1.fa --ref-fp db2.fa ..)\n"
+                   "default is: %s\n" % neg_db)
+@click.option('--neg-ref-db-fp', required=False, multiple=True,
+              type=click.Path(resolve_path=True, readable=True, exists=False,
+                              file_okay=True),
+              help="The indexed  version of the negative filtering database. "
+                   "If not supllied, deblur will index the database."
```
mortonjt: `supllied` -> `supplied`

deblur/test/test_workflow.py
```diff
@@ -186,6 +188,72 @@ def test_dereplicate_seqs(self):
...
+        """ Test filter_minreads_samples_from_table() function
+        for removal of samples with small number of reads
+        using the s4 dataset biom table
+        """
+        input_biom_file = join(self.test_data_dir, 'final.s4.biom')
+        table = load_table(input_biom_file)
+
+        # test basic filtering with 0 reads does not remove ok sample
+        new_table = filter_minreads_samples_from_table(table)
+        self.assertEqual(new_table.shape[1], 1)
```
antgonza: Is there a reason you test the shape of the matrices vs. testing the values in the matrices?

deblur/test/test_workflow.py
```diff
@@ -186,6 +188,72 @@ def test_dereplicate_seqs(self):
...
+        input_fasta_file = join(self.test_data_dir, 'final.s4.seqs.fa')
+        remove_artifacts_from_biom_table(input_biom_file,
+                                         input_fasta_file,
+                                         [pos_ref_fp],
+                                         self.working_dir,
+                                         pos_ref_db_fp)
+
+        origfilename = join(self.test_data_dir, 'simset.s2.fasta')
+        trim_length = 150
+        orig_seqs = [item[1] for item in sequence_generator(origfilename)]
```
antgonza: Could you add a comment on the magic number 1?

deblur/workflow.py
```diff
@@ -207,6 +208,101 @@ def build_index_sortmerna(ref_fp, working_dir):
...
+                                     negate=False, threads=threads,
+                                     verbose=verbose,
+                                     sim_thresh=sim_thresh,
+                                     coverage_thresh=coverage_thresh)
+    logger.debug('removed artifacts from sequences input %s'
+                 ' to output %s' % (fasta_filename, clean_fp))
+
+    # read the clean fasta file
+    good_seqs = set()
+    for chead, cseq in sequence_generator(clean_fp):
```
antgonza: No necessary to change but all this block could be:
`good_seqs = set([s for h, s in sequence_generator(clean_fp)])`

wasade: or slightly shorter: `good_seqs = {s for _, s in sequence_generator(clean_fp)}`

antgonza: The only "technical" issue, which we could also ignore, with @wasade's version is that if clean_fp is empty it will create an empty dict, not an empty set. As you can imagine this is completely out of the scope of this PR 😃 .

wasade: nope, empty `set`:

```python
>>> x = {i for i in []}
>>> x
set()
>>> type(x)
<class 'set'>
```

mortonjt: I think @wasade 's version is fine.

antgonza:
@amnona, could you address the comments? They should be pretty fast. Thanks!

antgonza:
@amnona, friendly ping. Remember that this is blocking other development ...

amnona:
Sorry. Been working intensively on the manuscript... Will try to address
this weekend :)

deblur/workflow.py
amnona: addressed :)

deblur/workflow.py
amnona: addressed :)

deblur/test/test_workflow.py
amnona: since we have 1 sample in the test table, enough to see if it is filtered away or not...

deblur/workflow.py
amnona: Who do i like better... Mom or Dad?
went with the majority - took @wasade 's version :)

amnona:
ok i think i addressed everything...

PR #94, opened 2016-10-17
fix: also work with a directory of .fastq files

amnona:


deblur/workflow.py
```diff
@@ -576,6 +577,10 @@ def get_files_for_table(input_dir,
     file_end : string
         the ending of all the fasta files to be added to the table
         (default '.fasta.trim.derep.no_artifacts.msa.deblur.no_chimeras')
+    num_extra_name_chars : int (optional)
+        the number of additional characters to remove from the end of the
+        file name to get the sample name
+        default is len('.fasta')=6
```
mortonjt: Is this necessary?  Can't we just do a split('.') to remove the ending?

josenavas: Adding to @mortonjt can you check `os.path.splitext`?

amnona: i think the problem is we don't know the sample name does not contain '.' in it.
i.e. say we start with a samples called "X.Y"
the reads file will be "X.Y.fasta"
and then the final deblurred file will be called "X.Y.fasta.trim.derep.ZZZ.nochimers"
we can use rsplit instead but i think this is more clear?

antgonza: I think what they are suggesting is something like: 
```python
In [1]: cfile = 'my_name.is_this.thing.fasta.trim.derep.no_artifacts'
In [2]: '.'.join(basename(cfile).split('.')[:-4])
Out[2]: 'my_name.is_this.thing'
```

josenavas: I think asking the user to count characters is not the best way to move forward. If we know that at this point the files are not compressed, we can safely used `os.path.splitext` which will remove the extension. (it will be similar to a rsplit, but safer, you can check the documentation: https://docs.python.org/2/library/os.path.html#os.path.splitext)

josenavas: @amnona can you provide a specific example on where this is useful? You already have the `file_end` to remove the end of the file name (including all extensions) so I'm really confused on why you need to remove that part.

mortonjt: 👍 for `os.path.splitext`

mortonjt:
I think this is good to go.  @antgonza @josenavas want to give this PR a blessing?

deblur/test/data/testmerge.fastq.trim.derep.no_artifacts.msa.deblur.no_chimeras
```diff
@@ -0,0 +1,20 @@
+>810205;size=1
```
josenavas: This file is called FASTQ but it is still a FASTA? I think it will make more sense to do the test with an actual FASTQ, just to make sure that it works as expected.

amnona: all deblurred sample files are fasta. the .fastq just means it started as a .fastq, but it turns into a fasta file after the .trim step

PR #96, opened 2016-10-26
try to fix the filtering bug

amnona:


josenavas:
Please - hold on merging this - i'm going to test first if it fixes the issue

wasade:
Testing to see if the `os.path.exists` etc calls can be omitted as the docstring for `sequence_generator` is not accurate with respect to the expectations of `input_fp`

antgonza:
BTW this solve the issue, thanks! Could you fix the coveralls so we can merge?

josenavas:
@wasade, do you think that you can add a test in which the code inside [this ](https://github.com/biocore/deblur/pull/96/files#diff-352208b099fc0ad645c237f32699335bR274) if statement is exercised?

@antgonza how strong do you feel about that?

antgonza:
I don't feel strongly about it but coveralls does ... it failed.

wasade:
...if this is blocking or critical, we could just merge and/or someone else can push in a test to exercise this branch

mortonjt:
Considering that this is just testing log statements and we have not been testing log statements in the past, I think that this is not in the scope this PR.  I just raised an issue [here](https://github.com/biocore/deblur/issues/97) so consider this resolved.

PR #98, opened 2016-11-02
BUGFIX: default negative database was set incorrectly

amnona:
Please merge fast :)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,16 @@
 # deblur changelog
 
+## Version 0.1.3-dev
+### Features
+
+### Backward-incompatible changes [stable]
+
+### Performance enhancements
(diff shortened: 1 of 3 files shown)


mortonjt:
whoa.  how did we miss that haha

antgonza:
Wow!! I'm also impress that the tests didn't catch that and/or that tests don't need to change ... Fine to merge after we understand why tests don't need to change or they are changed.

wasade:
Given this, can a regression test be added?

mortonjt:
Agree. We should add a test to make sure this doesn't happen again

wasade:
Please add a detailed mention of the problem into the CHANGELOG, and this should trigger a minor version release

PR #111, opened 2016-11-05
fix bug in parallel positive mode 

amnona:
(click supplies empty tuple instead of None)


josenavas:
Can we please add a test?

amnona:
added the test to the test_script.
can someone merge this (don't want users to install a version with a bug...)

josenavas:
👍 @antgonza wanna review/merge?

PR #124, opened 2016-12-07
skip trimming option

mortonjt:
This adds an option to ignore trimming.

So you should be able to specify a flag to completely ignore trimming.

Here a test case is added.  In addition, I have tested out the CLI on some of the test datasets.
Note that this is a little tricky to test, so any suggestions on improving this will be welcome.  

But I believe that this should cover the gist of it.


wasade:
thanks! there are some failures btw

cc @antgonza

mortonjt:
A fix for the file path is coming in - its passing on my system.  But this is ready for review.

deblur/workflow.py
```diff
@@ -770,9 +772,15 @@ def launch_workflow(seqs_fp, working_dir, mean_error, error_dist,
     # Step 1: Trim sequences to specified length
     output_trim_fp = join(working_dir, "%s.trim" % basename(seqs_fp))
     with open(output_trim_fp, 'w') as out_f:
-        for label, seq in trim_seqs(
-                input_seqs=sequence_generator(seqs_fp), trim_len=trim_length):
-            out_f.write(">%s\n%s\n" % (label, seq))
+        if skip_trim:
+            for label, seq in sequence_generator(seqs_fp):
```
antgonza: This is basically duplicating the file, right? Can we do something smarter?

josenavas: What about creating a simple [symlink](https://docs.python.org/2/library/os.html#os.symlink)?

mortonjt: That's cool!  I didn't know about that.

mortonjt:
Ok.  I think this is ready to merge!

mortonjt:
Thanks!

PR #130, opened 2017-02-01
CLI default values

amnona:
Updated --min-reads default to 10
Removed --trim-length default value and made it required
Changed --negate to be default True
Updated doc and Chengelog


josenavas:
--trim-length should not be required - as we are using it in Qiita is optional and we have the parameter --skip-trimming to not trim the data.

amnona:
We don't want to have a default value for the trim, as we saw users are not
aware of this parameter and don't understand the default length trimming.
I think Qiita can supply --trim-length 0 if --skip-trimming is supplied

amnona:
i think this complicates the --help option since then we don't set
--trim-length to required=True, so the --help will not show it as
required...
Any ideas how to solve?

antgonza:
In the past for other tools we just make both optional and clearly state in the help text of the parameters that one of them but not both are required.

wasade:
Requiring trim length is consistent with DADA2. 

What about removing the `--skip-trimming` option, interpret the value of `-1` for the trim length as meaning to skip trimming, and state this unambiguously in the help text?

README.md
```diff
@@ -24,7 +24,19 @@ At the moment, the install is a two stage process as we do not currently have de
...
+for mac OS X:
+
+```
+conda install -c biocore SortMeRNA==2.0
+```
+
+for other operating systems:
+
+```
+conda install -c bioconda SortMeRNA==2.0
```
wasade: This works for OSX too, why the two different options here? The build was [fixed](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/sortmerna/2.0/meta.yaml) a few months ago

README.md
```diff
@@ -53,7 +65,7 @@ and use the split/seqs.fna as the input to the deblur workflow.
 
 Important options
 =================
-- The sequence read length can be specified by the ```-t NNN``` flag, where NNN denotes the length all sequences will be trimmed to (default=150). Note that all reads shorter than this length will be discarded.
+- The sequence trim length is specified by the ```-t NNN``` flag, where NNN denotes the length all sequences will be trimmed to. Note that all reads shorter than this length will be discarded.
```
wasade: Can you add a single sentence here as to why trimming is necessary?

amnona:
using trim length -1 for no trimming sounds a good option to me :)
Jose/Antonio - what do you think?

josenavas:
I think it's acceptable. It will require some changes in the Qiita plugin but I think it will not be too bad. @antgonza ?

README.md
```diff
@@ -64,17 +76,17 @@ deblur workflow --help
...
 
-```
-deblur workflow --seqs-fp all_samples.fna --output-dir output -n
-```
+By default, deblur uses negative mode filtering to remove known artifact (i.e. PhiX and Adapter sequences) prior to denoising. The output of deblur contains two files: final.biom, which includes all sOTUs, and final.only16s.biom, which contains the output of positive filtering of the sOTUs (only sOTUs similar to 16S sequences).
+
+Minimal Reads Filtering
+=======================
+Deblur runs on each sample independently. However, sometimes there is also additional information based on the total number of times an sOTU is observed in all samples (e.g. an sOTU which is observed only in one sample at low read count is more likely to be pcr/read error as opposed to an sOTU present in many samples). The --min-reads option allows to use this information by removing sOTUs with a total read count (across all samples) lower than the given threshold. The default value is set to 10, and should be useful for most cases. However, if all samples in an experiment are expected not to contain the same bacteria, --min-reads can be set to 0 to skip this final filtering step.
```
wasade: I don't understand the statement "However, if all samples in an experiment are expected not to contain the same bacteria, --min-reads can be set to 0 to skip this final filtering step." Can this be restated?

scripts/deblur
```diff
@@ -198,11 +198,13 @@ def dereplicate(seqs_fp, output_fp, min_size, log_level, log_file):
                    "database. For multiple databases, the order "
                    "must follow that of --ref-fp, for example, "
                    "--ref-db-fp db1.idx --ref-db-fp db2.idx ..")
-@click.option('--negate', '-n', required=False, default=False, is_flag=True,
-              show_default=True, type=bool,
-              help="Discard all sequences aligning to the database "
-                   "passed via --ref-fp option")
-@click.option('--threads-per-sample', '-t', required=False, type=int,
+@click.option('--negate/--only-pos', '-n/', required=False, default=True, is_flag=True,
```
wasade: What about just removing this option and always doing both? I don't understand the usecase of just doing one or the other.

README.md
wasade: Also, I seem to remember you had generated a figure at one point to support the use of min-reads=10. Is that the case, and would it make sense to refer to here?

scripts/deblur
```diff
@@ -409,40 +411,43 @@ def build_biom_table(seqs_fp, output_biom_fp, min_reads, file_type, log_level,
...
+              show_default=True, help="Sequence trim length")
 @click.option('--pos-ref-fp', required=False, multiple=True,
               default=[], show_default=False,
               type=click.Path(resolve_path=True, readable=True, exists=True,
                               file_okay=True),
               help="Positive (16S) filtering database. "
                    "Keep all sequences aligning to this FASTA file "
-                   "(for multiple databases, use "
-                   "--ref-fp db1.fa --ref-fp db2.fa ..)\n"
+                   "in the output final.only16s.biom file. (for multiple "
```
wasade: I really think we should revisit the names of these files...

scripts/deblur
amnona: i think it's a good idea :)
now that positive runs by default after creating the biom table, should be enough.
(there is an edge case where you are only interested in 16s then running positive mode in the beginning - before deblur - will be more efficient compared to running negative mode, deblurring and then positive filtering)
but i think maybe simplicity is more important than this edge case?
so let's do it !

README.md
```diff
@@ -64,17 +76,17 @@ deblur workflow --help
...
-```
+- negative mode - removes [known artifact sequences](deblur/support_files/artifacts.fa) (i.e. PhiX and Adapter sequences).
 
-Negative filtering can be selected using the '-n' flag. This causes deblurring to keep all sequences except for [known artifact sequences](deblur/support_files/artifacts.fa) (i.e. PhiX and Adapter sequences), so other non-16S sequences are retained. For example:
+- positive mode - keeps only sequences similar to [known 16S sequences](deblur/support_files/88_otus.fasta) (GreenGenes 88%)
 
-```
-deblur workflow --seqs-fp all_samples.fna --output-dir output -n
-```
+By default, deblur uses negative mode filtering to remove known artifact (i.e. PhiX and Adapter sequences) prior to denoising. The output of deblur contains two files: final.biom, which includes all sOTUs, and final.only16s.biom, which contains the output of positive filtering of the sOTUs (only sOTUs similar to 16S sequences).
```
amnona: i agree. we can use new names. maybe all.biom and positive-filtered.biom ?
what do other people think

wasade: not "unfiltered.biom"?

scripts/deblur
wasade: ya, agree on the simplicity

@antgonza @josenavas fyi, this change would trigger a change in the qiita plugin

README.md
wasade: @antgonza @josenavas this also would trigger a change on qiita side

antgonza:
@wasade / @amnona / @josenavas  I think those changes are fine and we can simply rename the files in the system when we decide to redeploy this plugin or just leave as is. The changes in the pipeline are not algorithmic so we can avoid reprocessing.

wasade:
Excellent, thank you @antgonza 

@amnona ETA on a commit?

README.md
amnona: cool :)
fixed.

README.md
amnona: added

README.md
```diff
@@ -64,17 +76,17 @@ deblur workflow --help
...
-By default, deblur uses positive filtering, keeping only 16S sequences (based on homology to [Greengenes](http://greengenes.secondgenome.com/) [88% representative set](deblur/support_files/88_otus.fasta)). For example:
+Deblur can use two types of filtering on the sequences:
 
-```
-deblur workflow --seqs-fp all_samples.fna --output-dir output
-```
+- negative mode - removes [known artifact sequences](deblur/support_files/artifacts.fa) (i.e. PhiX and Adapter sequences).
 
-Negative filtering can be selected using the '-n' flag. This causes deblurring to keep all sequences except for [known artifact sequences](deblur/support_files/artifacts.fa) (i.e. PhiX and Adapter sequences), so other non-16S sequences are retained. For example:
+- positive mode - keeps only sequences similar to [known 16S sequences](deblur/support_files/88_otus.fasta) (GreenGenes 88%)
```
amnona: :)

README.md
amnona: tried to make it more clear. let me know if it makes sense.

README.md
```diff
@@ -64,17 +76,17 @@ deblur workflow --help
 
 Positive and Negative Filtering
 ===============================
-By default, deblur uses positive filtering, keeping only 16S sequences (based on homology to [Greengenes](http://greengenes.secondgenome.com/) [88% representative set](deblur/support_files/88_otus.fasta)). For example:
+Deblur can use two types of filtering on the sequences:
```
amnona: added

scripts/deblur
amnona: changed to:
all.biom - all deblurred sequences
positive-match.biom - only sequences matching positive database (16S)
positive-nonmatch.biom - only the non 16S sequences

amnona:
ok
addressed all comments (except the figure for the min-reads ... need to see how to do it nicely)
lets merge

README.md
```diff
@@ -35,12 +35,18 @@ pip install deblur
 Example usage
 =============
 
-The input to deblur workflow is a directory of fasta files (1 per sample) or a demultiplexed FASTA or FASTQ file. The output is a biom table with sequences as the OTU ids (final.biom in the output directory).
+The input to deblur workflow is a directory of fasta files (1 per sample) or a demultiplexed FASTA or FASTQ file. The output are three biom tables (in the output directory) with sequences as the OTU ids:
 
-The simple use case just specifies the input fasta file (or directory) and output directory name:
+- all.biom : contains all deblurred reads.
+
+- positive-match.biom : contains only deblurred reads matching the positive filtering database (default is only 16S reads)
```
wasade: what about "reference-hit.biom" and "reference-misshit.biom"? issue with match is that it implies it is a strong hit, however the whole point is to get things which look approximately like the reference. Second, it will be a rare thing for someone to change up the negative reference, so I think it's easier for users to just omit, are label it as the "artifact-reference" or something.

wasade: @antgonza @josenavas @mortonjt @tanaes, can you comment please on what would be intuitive as filenames?

README.md
```diff
@@ -64,17 +70,19 @@ deblur workflow --help
...
-By default, deblur uses positive filtering, keeping only 16S sequences (based on homology to [Greengenes](http://greengenes.secondgenome.com/) [88% representative set](deblur/support_files/88_otus.fasta)). For example:
+Deblur uses two types of filtering on the sequences:
 
-```
-deblur workflow --seqs-fp all_samples.fna --output-dir output
-```
+- negative mode - removes [known artifact sequences](deblur/support_files/artifacts.fa) (i.e. sequences aligning to PhiX or Adapter with >=95% identity and coverage).
 
-Negative filtering can be selected using the '-n' flag. This causes deblurring to keep all sequences except for [known artifact sequences](deblur/support_files/artifacts.fa) (i.e. PhiX and Adapter sequences), so other non-16S sequences are retained. For example:
+- positive mode - keeps only sequences similar to [known 16S sequences](deblur/support_files/88_otus.fasta) (e-value<=10 for Greengenes 88% rep. set). Note that deblur also outputs a biom table without this positive filtering step.
```
wasade: an e-value of 10??

wasade: note, "by default using a database of known 16S sequences" or something to that effect

deblur/test/test_workflow.py
```diff
@@ -748,9 +747,9 @@ def test_launch_workflow_skip_trim(self):
         indel_prob = 0.01
         indel_max = 3
         min_size = 2
-        negate = False
+        # negate = False
```
wasade: remove please

deblur/test/test_workflow.py
```diff
@@ -788,7 +785,7 @@ def test_launch_workflow_incorrect_trim(self):
...
```
wasade: remove please

wasade:
@amnona ping

README.md
amnona: Yup. sortmerna e<=10 is what Jenya recommended, and also works well in the cases we tested.

amnona: rephrased

deblur/test/test_workflow.py
amnona: fixed

deblur/test/test_workflow.py
amnona: fixed

amnona:
ok
all should be good to go
output files are now named : all.biom, reference-hit.biom and reference-non-hit.biom

README.md
amnona: renamed to all.biom, reference-hit.biom and reference-non-hit.biom

Issue #104, opened 2016-11-03
add --version option


amnona:
deblur --version should return the current version number

amnona:
don't know how to do it elegantly... anyone?

wasade:
An example is [here](https://github.com/biocore/biom-format/blob/master/biom/cli/__init__.py#L18)

amnona:
cool. added.

PR #131, opened 2017-02-07
Milestone1

amnona:
fixes for milestone1 issues


deblur/workflow.py
```diff
@@ -783,29 +790,33 @@ def launch_workflow(seqs_fp, working_dir, mean_error, error_dist,
...
-    alignment = multiple_sequence_alignment(seqs_fp=output_artif_fp,
-                                            threads=threads_per_sample)
-    if not alignment:
-        warnings.warn('Problem performing multiple sequence alignment '
-                      'on file %s' % seqs_fp, UserWarning)
-        logger.warning('msa failed. aborting')
-        return None
-
+    if num_seqs_left > 1:
+            output_msa_fp = join(working_dir,
```
wasade: please make sure the code is intended properly

wasade:
Please note what issues are included here, and refer to the issues appropriately using # so that they get marked.

deblur/workflow.py
```diff
@@ -434,7 +441,7 @@ def op(x): return x in aligned_seq_ids
                 badseqs += 1
     logger.info('total sequences %d, passing sequences %d, '
                 'failing sequences %d' % (totalseqs, okseqs, badseqs))
-    return output_fp
+    return output_fp, okseqs
```
wasade: to verify, were all instances of the calls to this method updated with respect to the API change?

deblur/workflow.py
amnona: done

deblur/workflow.py
amnona: yup :)

deblur/workflow.py
```diff
@@ -783,29 +790,33 @@ def launch_workflow(seqs_fp, working_dir, mean_error, error_dist,
...
     # Step 3: Remove artifacts
-    output_artif_fp = remove_artifacts_seqs(seqs_fp=output_derep_fp,
-                                            ref_fp=ref_fp,
-                                            working_dir=working_dir,
-                                            ref_db_fp=ref_db_fp,
-                                            negate=True,
-                                            threads=threads_per_sample,
-                                            sim_thresh=sim_thresh)
+    output_artif_fp, num_seqs_left = remove_artifacts_seqs(seqs_fp=output_derep_fp,
+                                                           ref_fp=ref_fp,
```
wasade: how come this passed as a tuple `(ref_fp, )` like the other calls to `remove_artifacts_seqs` edited above?

deblur/workflow.py
```diff
@@ -783,29 +790,33 @@ def launch_workflow(seqs_fp, working_dir, mean_error, error_dist,
...
+                             "%s.msa" % basename(output_artif_fp))
+        alignment = multiple_sequence_alignment(seqs_fp=output_artif_fp,
+                                                threads=threads_per_sample)
+        if not alignment:
+            warnings.warn('Problem performing multiple sequence alignment '
+                          'on file %s' % seqs_fp, UserWarning)
+            logger.warning('msa failed. aborting')
+            return None
+    else:
+        # only one sequence after remove artifacts (but could be many reads)
```
wasade: it's also possible that `num_seqs_left == 0`, right?

scripts/deblur
```diff
@@ -647,6 +650,8 @@ def workflow(seqs_fp, output_dir, pos_ref_fp, pos_ref_db_fp,
         for f in files_to_remove:
             remove(f)
         rmtree(working_dir)
+        if created_split_dir is not None:
```
wasade: it is not assured that `created_split_dir` exists as a variable at this point. it is assumed to be the case with the current structure of the logic, but this is a bug waiting to happen if the parallel logic ever changes

wasade:
Please note what issues are included here, and refer to the issues appropriately using # so that they get marked. For instance, it appears that #104 is resolved in this PR. By mentioning it, as I just did, the issue gets a reference making it easy to know what issues are associated with what changes, and so we can make sure that issues get closed out as they are resolved.

deblur/workflow.py
amnona: sorry, don't understand the comment... can you explain more?

wasade: See [here](https://github.com/biocore/deblur/pull/131/files#diff-352208b099fc0ad645c237f32699335bR794) compared to [here](https://github.com/biocore/deblur/pull/131/files#diff-bd2728968a097d66eb37926bfb6ba6fcR316)

deblur/workflow.py
amnona: the pipeline should take care of it by failing the next step
but just to be on the safe/clear side, added explicitly num_seqs_left == 0 condition

scripts/deblur
amnona: should not be undefined in current implementation but agree could complicate things later if logic is changes. moved the initialization closer to beginning of pipeline.

Issue #133, opened 2017-02-08
Undescribed output files


wasade:
There are a few output files which are not described in the readme I/O files section. Specifically:

- all.seqs.fa
- all.seqs.fa.sortmerna.blast
- all.seqs.fa.no_artifacts

wasade:
I believe this is addressed, and that .sortmerna.blast and .no_artifacts are no longer part of the output.

PR #135, opened 2017-02-09
address #133

amnona:
remove 2 tmp files
add the all.seqs.fa to doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,11 @@ The input to Deblur workflow is a directory of FASTA or FASTQ files (1 per sampl
 
 - reference-hit.biom : contains only Deblurred reads matching the positive filtering database. By default, a reference composed of 16S sequences is used, and this resulting table will contain only those reads which recruit at a coarse level to it will be retained. Reads are also filtered against the negative reference, which by default will remove any read which appears to be PhiX or adapter.
 
+- reference-hit.seqs.fa : a fasta file containing all the sequences in reference-hit.biom
+
 - reference-non-hit.biom : contains only Deblurred reads that did not align to the positive filtering database. Negative filtering is also appied to this table, so by default, PhiX and adapter are removed.
+
+- reference-non-hit.seqs.fa : a fasta file containing all the sequences in reference-non-hit.biom
 
 - all.biom : contains all Deblurred reads. This file represents the union of the "reference-hit.biom" and "reference-non-hit.biom" tables.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -61,6 +61,8 @@ The input to Deblur workflow is a directory of FASTA or FASTQ files (1 per sampl
 
 - all.biom : contains all Deblurred reads. This file represents the union of the "reference-hit.biom" and "reference-non-hit.biom" tables.
 
+- all.seqs.fa : a fasta file containing all the sequences in all.biom
+
```
wasade: Please also describe the contents of `all.seqs.fa.sortmerna.blast` and `all.seqs.fa.no_artifacts`

scripts/deblur
```diff
@@ -649,12 +649,14 @@ def workflow(seqs_fp, output_dir, pos_ref_fp, pos_ref_db_fp,
     create_otu_table(output_fp, samples,
                      outputfasta_fp=outputfasta_fp, minreads=min_reads)
 
-    # if negative mode, also create a biom table with
+    # also create a biom table with
     # only 16s sequences and with only non-16s sequences
```
wasade: not 16s specific

README.md
amnona: all.seqs.fa.sortmerna.blast and all.seqs.fa.no_artifacts are gone :)
they are now added to the tmp_files and deleted at the end of the deblur pipeline unless --keep-tmp-files is selected
so no need to describe them :)

josenavas: Does this mean that the only output fasta that deblur will generate now is the all.seqs.fa? If so, I disagree with the change as the file is useful for subsequent analysis, like generate a tree

wasade: I agree. Keeping a file of the sequences represented in the BIOM table that will be used for downstream analysis, in the vast majority of cases, makes sense to produce.

antgonza: Agree.

josenavas:
Just the comment about removing the sequence files - having the sequence file paired with the biom table is important for downstream analysis. Otherwise this looks good.

README.md
wasade: So it turns out, neither of the existing two sequence files corresponded to `reference-hit.biom` (formerly the only-16s table). 

@amnona, what we need are:

- a FASTA corresponding to `reference-hit.biom`
- a FASTA corresponding to `reference-non-hit.biom`

deblur/workflow.py
```diff
@@ -262,6 +262,25 @@ def filter_minreads_samples_from_table(table, minreads=1, inplace=True):
     return table
 
 
+def fasta_from_biom(table, fasta_file_name):
+    '''Save squences from a biom table to a fasta file
```
wasade: squences -> sequences

PR #140, opened 2017-02-23
Thresholds

### LINKED ISSUE #139: CRITICAL: similarity and coverage thresholds are not used in positive mode
Despite the [docs](https://github.com/biocore/deblur/blob/master/deblur/workflow.py#L391-L400), they are not [used](https://github.com/biocore/deblur/blob/master/deblur/workflow.py#L464-L466). 

Even had they been, a coverage threshold of 30% is massively permissive and allows obviously non-reference sequences to filter through. 

Issuing a PR in a moment.

wasade:
Fixes #139. 

A micro release will need to be cut after merge. @mortonjt @RNAer , just FYI as this will need to be reflected in conda, and will impact q2-deblur.

@antgonza @josenavas available for review? 

Do not merge until after a consensus has been met on this PR.

### DIFF
--- a/deblur/workflow.py
+++ b/deblur/workflow.py
@@ -396,7 +396,7 @@ def remove_artifacts_seqs(seqs_fp,
     coverage_thresh: float, optional
         The minimal coverage threshold (between 0 and 1)
         for alignments for keeping the sequence
-        if None, the default values used are 0.6 for negate=False,
+        if None, the default values used are 0.5 for negate=False,
         0.95 for negate=True
 
     Returns
@@ -419,7 +419,7 @@ def remove_artifacts_seqs(seqs_fp,
         if negate:
             coverage_thresh = 0.95 * 100
         else:
-            coverage_thresh = 0.6 * 100
+            coverage_thresh = 0.5 * 100
 
     if sim_thresh is None:
         if negate:
--- a/deblur/workflow.py
+++ b/deblur/workflow.py
@@ -396,7 +396,7 @@ def remove_artifacts_seqs(seqs_fp,
     coverage_thresh: float, optional
         The minimal coverage threshold (between 0 and 1)
         for alignments for keeping the sequence
-        if Nonr, the default values used are 0.3 for negate=False,
+        if None, the default values used are 0.6 for negate=False,
         0.95 for negate=True
 
     Returns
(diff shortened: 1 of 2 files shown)


deblur/workflow.py
```diff
@@ -456,14 +456,9 @@ def remove_artifacts_seqs(seqs_fp,
...
-                if negate:
-                    if (float(line[2]) >= sim_thresh*100) and \
-                       (float(line[13]) >= coverage_thresh*100):
-                        aligned_seq_ids.add(line[0])
-                else:
-                    if float(line[10]) <= 10:
-                        aligned_seq_ids.add(line[0])
+                if (float(line[2]) >= sim_thresh) and \
+                   (float(line[13]) >= coverage_thresh):
+                    aligned_seq_ids.add(line[0])
```
amnona: We are not using the E-value (position 10) criteria anymore?

deblur/workflow.py
```diff
@@ -417,15 +417,15 @@ def remove_artifacts_seqs(seqs_fp,
 
     if coverage_thresh is None:
         if negate:
-            coverage_thresh = 0.95
+            coverage_thresh = 0.95 * 100
         else:
-            coverage_thresh = 0.3
+            coverage_thresh = 0.6 * 100
```
amnona: Do we know 60% coverage/65% similarity are good for identifying 16S based on gg 88%?

deblur/workflow.py
wasade: Jenya advised using the default of 1

deblur/workflow.py
wasade: If a sequence doesn't align at 60% of its length to greengenes, do you think it is likely to be 16S given how conserved it is in many regions?

amnona: the problem is that sometimes it can go for a shorter alignment with higher identity compared to longer alignment with lower identity (i.e. maybe 50% coverage 100% identity instead of 100% coverage 80% identity)... I'm not sure how it chooses...

wasade:
Thresholds are now based off cross validation analysis against Greengenes 88%.

wasade:
Is this good to go? @antgonza gave support on the thresholds [here](https://github.com/biocore/deblur/issues/139#issuecomment-282271473)

amnona:
👍 
looks good to me
(there's always a type1/type2 balance, but it looks like a good compromise)

wasade:
Thanks!

@mortonjt, do you have bandwidth to do the 1.0.1 release? It would be great
to follow the checklist we did for 1.0.0 to make sure we cover all our
bases. I'll be able to do reviews today. Let me know if you don't, and I'll
figure out another solution

mortonjt:
Is it ready for release?  Sure I can do it.

mortonjt:
I don't see that the version number has been updated ... is there a checklist for the 1.0.1 release?

PR #146, opened 2017-03-16
replace e-value with bit-score filtering

amnona:
for reference/non-reference hits

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,16 @@
 # deblur changelog
 
+## Version 1.0.2
+
+### Features
+
+### Backward-incompatible changes [stable]
+
+### Performance enhancements
+
+### Bug fixes
+E-value thresholding removed in favor of a bitscore threshold which is scaled based on the length of the query sequence. This is to make each query sequence independent of each other on assessment against the positive filtering database.
+
 ## Version 1.0.1
 
 Official version 1.0.1 Released on 28 February 2017.
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,7 +9,7 @@
 ### Performance enhancements
 
 ### Bug fixes
-E-value thresholding removed in favor of a bitscore threshold which is scaled based on the length of the query sequence. This is to make each query sequence independent of each other on assessment against the positive filtering database.
+E-value thresholding removed in favor of a bitscore threshold which is scaled based on the length of the query sequence. This is to make each query sequence independent of each other on assessment against the positive filtering database. Please see [PR #146](https://github.com/biocore/deblur/pull/146) for further details.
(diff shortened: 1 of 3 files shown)


wasade:
Thanks! Will pull down in a second

wasade:
Further note, this is additional analysis of the sortmerna benchmarking (see [here](https://github.com/biocore/deblur/issues/139#issuecomment-282143373)).

wasade:
@amnona, small note, can you please bump `__init__.__version__` to `1.0.1-dev0` please? That should have been done already but slipped through on the release.

mortonjt:
👍

deblur/workflow.py
```diff
@@ -427,6 +427,9 @@ def remove_artifacts_seqs(seqs_fp,
         else:
             sim_thresh = 0.65 * 100
 
+    # the minimal average bitscore per nucleotide
+    bitscore_thresh = 0.65
```
ekopylova: can this be a parameter a user can change?

amnona: I think it's better not to have it as a CLI parameter? We have lot's of parameters already, and this is just a convenience function for running SortMeRNA - so if a user wants finer control as to what's considered 16S/not 16S her can run SortMeRNA on all.biom with whatever database/parameters he wants and get the same thing...
But can also add the CLI parameter - not strong opinion - what do others think?

mortonjt: Overall, I think it is a better idea to expose this as a parameter.  

1. If the user wants to tune this threshold, they can
2. Less maintenance burden for us, so that we decide that we don't like this threshold, we don't have to alter the underlying code.

ChangeLog.md
```diff
@@ -1,5 +1,16 @@
...
+## Version 1.0.2
+
+### Features
+
+### Backward-incompatible changes [stable]
+
+### Performance enhancements
+
+### Bug fixes
+E-value thresholding removed in favor of a bitscore threshold which is scaled based on the length of the query sequence. This is to make each query sequence independent of each other on assessment against the positive filtering database.
```
wasade: Please include a link to this pull request

deblur/workflow.py
wasade: There are a lot of parameters which are not exposed. In general, I agree with you @mortonjt and @ekopylova that it would be useful to facilitate exploration in the future but I think it may be easier to do this in a single shot as it would arguably make sense to package the user accessible parameters into an object. In other words, let's not do this now -- can you create an issue about this though?

Also, @mortonjt, if we change it in the future, it's either inline or the default value, so we don't get away from altering code somewhere.

wasade:
Thanks!

Issue #158, opened 2017-10-13
update required vsearch version


gregcaporaso:
The vsearch version that is currently required (2.0.3) has [a bug](https://github.com/torognes/vsearch/issues/201) that is [preventing some functionality from being added to the `q2-vsearch` plugin](https://github.com/qiime2/q2-vsearch/issues/24). It would be helpful if deblur could use a more recent version of vsearch. The vsearch bug was reported to have been fixed in 2.1.1. (I don't think this vsearch bug impacts deblur.)

ElDeveloper:
Closed in #159 though I believe the conda recipe will need to be updated (after a new release is published).

wasade:
@ElDeveloper, can this issue be closed?

ElDeveloper:
I think so.

PR #159, opened 2017-10-17
BLD: setting vsearch req to >= 2.0.3, seems to work with 2.5.0 w/o issue

wasade:
Changelog changes were because entries were going not under 1.0.3-dev


wasade:
Fixes #158

ElDeveloper:
Should the README.md also be updated?

wasade:
Likely, good catch

PR #166, opened 2018-01-17
Update MacOS Matplotlib troubleshooting item

karlrl:
In recent versions of Matplotlib, the [error message is slightly different](https://github.com/matplotlib/matplotlib/commit/cfb196037d5a2cc2d2721ba94eeb5a5f92645ab6) than was in the README. This updated version truncates the message to the first two sentences (which have not changed), and cleans up the formatting slightly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,15 +96,16 @@ Deblur runs on each sample independently. However, sometimes there is also addit
 
 Troubleshooting
 ===============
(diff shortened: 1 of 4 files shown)


RNAer:
@karlrl  could you specify which exception expected in the change? and there are conflicts need to be fixed.

PR #167, opened 2018-01-19
Improve execution time

karlrl:
Using [Pysam](https://github.com/pysam-developers/pysam) (which wraps [htslib](http://www.htslib.org/)) speeds up the sequence trimming step by a good amount (~50x). This changeset switches from using `skbio.read()` to using `pysam.FastxFile()` and makes a few other changes to accommodate the `FastxFile()` requirement that the file be passed as a path (or stdin), not a Python file-like object.

As an example of the speedup, processing a ~50k FASTA file took 65s with `skbio.read()` and 1.4.s with `pysam.FastxFile()`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
 install:
   - conda create --yes -n env_name python=$PYTHON_VERSION pip numpy scipy nose flake8 h5py
   - source activate env_name
-  - conda install --yes -c bioconda "VSEARCH>=2.0.3" MAFFT=7.310 SortMeRNA=2.0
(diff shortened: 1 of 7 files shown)


karlrl:
The flake8 problems are addressed as part of #166. I'm happy to cherry-pick that here if the other PR is rejected.

wasade:
Sorry for the delayed response on this. While this could improve execution time for one aspect of the pipeline, the trimming component is not a dominant piece of the runtime. I'm concerned about adopting a complex dependency without stronger evidence of its need (and note trimming can be performed prior to deblur execution).

PR #168, opened 2018-01-23
fix lower case nucleotides

sjanssen2:
added a unit test to check that all deblurred fragments are all upper case


sjanssen2:
currently, it is supposed to fail, since this is not a fix only a test that detects the issue

sjanssen2:
I think what needs to be done is enforcing upper case for sequences in https://github.com/biocore/deblur/blob/a10f1b568b9467bdeb53f6a6d895a33e15cb5b94/deblur/workflow.py#L676

sjanssen2:
@antgonza @josenavas @mortonjt could you review please?

sjanssen2:
@tkosciol interested in a review :-)

deblur/workflow.py
```diff
@@ -720,13 +720,14 @@ def create_otu_table(output_fp, deblurred_list,
...
         for chead, cseq in sequence_generator(cfilename):
+            cseq = cseq.upper()
             if cseq not in seqdict:
                 seqdict[cseq] = len(seqlist)
                 seqlist.append(cseq)
             cseqidx = seqdict[cseq]
             cfreq = float(sizeregexp.search(chead).group(0))
             try:
-                obs[cseqidx, csampidx] = cfreq
+                obs[cseqidx, csampidx] += cfreq
```
mortonjt: Not clear to me what's going on here ...

sjanssen2: Amnon is composing the deblur count table from single counts per sample per feature. Since we might have the situation that we have counts for one sample for feature "ACUGA" with let's say 10 reads and also 5 reads for feature "ACUGa" we need to make sure that those reads are added up to be 15.
Thus, the same obs[cseqidx, csampidx] can be updated several times.

sjanssen2:
@antgonza forgot to change the title :-/

PR #177, opened 2018-07-14
BUGFIX: 1E9 as float, and fixing the unit tests

amnona:
BUGFIX
in recent python versions, 1E9 (which is used as the number of rows in the sparse matrix converted to biom) is treated as float rather than int, causing the allocation to fail
Also, unit testing expects in the output fasta to have "size=XXX;" whereas deblur now gives "size=XXX" (without the trailing ";"). Fixed the unit test.


amnona:
fixed the changelog :)
ready for merge

amnona:
fixed version to 1.0.4-dev and corresponding changelog

Issue #178, opened 2018-08-03
Indels aren't calculated correctly


MichaelRade:
Hi,
Indels aren't calculated correctly because `sub_seq_i` and `sub_seq_j` are identical.

deblurring.py
160 `sub_seq_i = seq_i.np_sequence[:length]`
161 `sub_seq_j = seq_i.np_sequence[:length]`

amnona:
Hi,
thanks for the feeback.
I didn't fully understand what you mean regarding the indel calculation problem.
can you please elaborate what exactly you think the problem is?
can you give an example where you think the indel calculation isn't correct (starting from the non-aligned sequences)?

thanks
Amnon

MichaelRade:
Hi,
for example:
`seq_i.sequence = ---GGAGGGT-----`
`seq_j.sequence = ---AGG-GCGG----`
`seq_i.np_sequence = [4 4 4 2 2 0 2 2 2 3 4 4 4 4 4]`
`seq_j.np_sequence = [4 4 4 0 2 2 4 2 1 2 2 4 4 4 4]`

In line 160/161 you say this:
`sub_seq_i = seq_i.np_sequence[:length]`
`sub_seq_j = seq_i.np_sequence[:length]`
 "sub_seq_i" and "sub_seq_j" assigned the same sequence. So "sub_seq_i == sub_seq_j"
Therefore:
`sub_seq_i  = [4 4 4 2 2 0 2 2 2 3 4 4 4 4 4]`
`sub_seq_j =  [4 4 4 2 2 0 2 2 2 3 4 4 4 4 4]`

in line 163 you say
`mask = (sub_seq_i != sub_seq_j)`
Because "sub_seq_i " and "sub_seq_j " are always equal all element in "mask" have the boolean false.

Consequently in line 165 "mut_is_indel " will be always an empty list
`mut_is_indel = np.logical_or(sub_seq_i[mask] == 4, sub_seq_j[mask] == 4)`

If you replace in line 161 **sub_seq_j = seq_i.np_sequence[:length]** with **sub_seq_j = seq_j.np_sequence[:length]** i think it will be work.

MichaelRade:
Hi,
thanks. II'm going to check this out. Regarding your question: I've decided to read the scripts if I want to use it for my analyses. Let's see how long I can keep that up :)

wasade:
Thank you @MichaelRade for reporting, and for a detailed example of the bug. I'm reopening the issue until a fix is in place.

[...]

PR #179, opened 2018-08-09
fix seq_i, seq_j type

amnona:
Fix the sub_seq_j = seq_i.np_sequence[:length] typo bug to seq_j.

Also, fix the excessive '-' at end of alignment being counted as mismatch when handling the indel.

Also fix the unit test, and validate the indel removal using the indel error rate and not the general mismatch error rate (by adding an indel containing sequence that should not be removed under the indel error rate, but will be removed with the mismatch error rate).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,10 @@
 # deblur changelog
 
+## Version 1.0.5-dev
+
+### Bug fixes
+* Fixed problem causing deblur to ignore the indel error uper bound and use the mismatch error upper bound instead. This caused deblur to use default error rate of 0.06/0.02 instead of 0.01 which is the default for indel.
+
 ## Version 1.0.4-dev
 
 ### Bug fixes
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,9 +5,6 @@
 ### Bug fixes
 * Fixed problem causing deblur to ignore the indel error uper bound and use the mismatch error upper bound instead. This caused deblur to use the mismatch depenent default error rate instead of the default indel error rate (which is constant for up to 3 indels). Also increased the default indel error rate to 0.02. See [issue #178](https://github.com/biocore/deblur/issues/178) for more details.
 
-## Version 1.0.4-dev
-
-### Bug fixes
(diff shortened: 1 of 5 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # deblur changelog
 
+## Version 1.0.5-dev
+
+### Bug fixes
+* Fixed problem causing deblur to ignore the indel error uper bound and use the mismatch error upper bound instead. This caused deblur to use default error rate of 0.06/0.02 instead of 0.01 which is the default for indel.
```
wasade: uper -> upper

Would it be possible to include a hyperlink to the issue? e.g., "See [issue #178](https://github.com/biocore/deblur/issues/178) for more detail"

ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # deblur changelog
 
+## Version 1.0.5-dev
```
wasade: I don't understand why the version number has been incremented, this should still fall under 1.0.4-dev right?

amnona:
Fixed comments
also, changed the default indel upper bound to 0.02 (instead of 0.01). This should make results more compatible with the previous version (since it used the mismatch bound which is 0.02 for 2 mismatches, which is what we get if the indel is not recognized).

ChangeLog.md
```diff
@@ -3,6 +3,8 @@
 ## Version 1.0.4-dev
 
 ### Bug fixes
+* Fixed problem causing deblur to ignore the indel error uper bound and use the mismatch error upper bound instead. This caused deblur to use the mismatch depenent default error rate instead of the default indel error rate (which is constant for up to 3 indels). Also increased the default indel error rate to 0.02. See [issue #178](https://github.com/biocore/deblur/issues/178) for more details.
```
wasade: Is there empirical data to support the change?

amnona:
Since all deblur testing and usage was done with the bug, updating the
indel probability to 0.02 makes the results more compatible.
Also, 0.02 is a more conservative upper bound.

antgonza:
after discussing with @wasade we are gonna merge and then fix in another pr ...

PR #184, opened 2018-09-10
fix #179 + changing versions for release

antgonza:


.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: python
 env:
   - PYTHON_VERSION=3.5
+  - PYTHON_VERSION=2.7
```
mortonjt: Just to clarify, this is intentional? We are supporting py2?

wasade:
:+1: pending tests

.travis.yml
antgonza: Well, I always thought we were but seems like due to skbio we are not ... I'm gonna delete from this PR ...

PR #197, opened 2019-06-19
[BUG] Bug fix in parameter name of remove-chimeras-denovo

TaskeHAMANO:
Although the second argument of remove-chimeras-denovo is an output directory name, it was expressed as an output file name. Fixed this.


wasade:
Thank you, @TaskeHAMANO, for these PRs! It looks like our travis configuration has grown slightly stale. If you have a second, would you be able to try removing "--process-dependency-links" from the `.travis.yml` file and see if that unsticks the build?

PR #210, opened 2022-05-31
finish mv to github actions

antgonza:
Moving code to github actions and settling on Python 3.6.


charles-cowart (changes requested):
I believe these are correct. However, it is not passing CI. Will approve once it passes.

charles-cowart (approved):
Everything looks good.

PR #219, opened 2023-08-07
MAINT: replace deprecated `.warn()` method

valentynbez:
The logging.Logger.warn() method was deprecated in Python 3.3 by issue https://github.com/python/cpython/issues/57444 and commit https://github.com/python/cpython/commit/04d5bc00a219860c69ea17eaa633d3ab9917409f. This method is not documented and emits a DeprecationWarning since Python 3.3.


wasade:
Thanks, @valentynbez! Is there any chance the lint issues could be addressed in this PR? I realize they're unrelated but I suspect are fast to do.

wasade:
Thank you!!

